PR #186, opened 2021-09-08
include locations in Core output if --pp_flags=annot specified

vzaliva:
According to command-line options description: "Pretty print flags [annot: include location and ISO annotations".
For Core files, this flag has no effect.  This PR fixes that.


kmemarian:
Do you also want the ISO annotations to be printed? The `WithLocations` module only prints locations, we could create another instance that does both.

vzaliva:
Yes, let me add ISO annotations for completeness

PR #196, opened 2022-05-18
Vaddr

vzaliva:
Added new CHERI-specific integer type `vaddr_t`.
It is always unsigned and has the range to hold memory addresses (e.g. 64 bits on int64).


vzaliva:
Sorry wrong upstram. let me re-create

PR #226, opened 2024-04-08
Dune version consistent

elefthei:
`cerberus-lib.opam` allows for `dune >= 3.15.0` so I made that the case for `cerberus.opam` as well. Otherwise `opam` takes the intersection of the two and force me to downgrade dune to `3.13.0`. Which I'd rather not do unless I absolutely have to.


kmemarian:
Oops, thanks for the PR.

Issue #233, opened 2024-04-25
CN crash when left shifting unsigned char or short


peterohanley:
`x.c`:
```
void a(void)
{
    unsigned char b = 0;
    b << 0;
}
```
```
% cn x.c 
[1/1]: a
cn: internal error, uncaught exception:
    File "backend/cn/check.ml", line 649, characters 5-11: Assertion failed
    Raised at Dune__exe__Check.check_pexpr in file "backend/cn/check.ml", line 649, characters 5-98
    Called from Dune__exe__Typing.pure in file "backend/cn/typing.ml", line 102, characters 22-25
    Called from Dune__exe__Typing.bind in file "backend/cn/typing.ml", line 78, characters 8-11
    Called from Dune__exe__Typing.pure in file "backend/cn/typing.ml", line 102, characters 22-25
    Called from Dune__exe__Typing.bind in file "backend/cn/typing.ml", line 78, characters 8-11
    Called from Dune__exe__Typing.bind in file "backend/cn/typing.ml", line 78, characters 8-11
    Called from Dune__exe__Typing.pure in file "backend/cn/typing.ml", line 102, characters 22-25
    Called from Dune__exe__Typing.bind in file "backend/cn/typing.ml", line 78, characters 8-11
    Called from Dune__exe__Typing.pure in file "backend/cn/typing.ml", line 102, characters 22-25
    Called from Dune__exe__Typing.bind in file "backend/cn/typing.ml", line 78, characters 8-11
    Called from Dune__exe__Typing.bind in file "backend/cn/typing.ml", line 78, characters 8-11
    Called from Dune__exe__Typing.bind in file "backend/cn/typing.ml", line 78, characters 8-11
    Called from Dune__exe__Typing.run in file "backend/cn/typing.ml", line 44, characters 16-19
    Called from Dune__exe__Main.main in file "backend/cn/main.ml", line 182, characters 20-87
    Called from Dune__exe__Main.main in file "backend/cn/main.ml", line 178, characters 9-1023
    Re-raised at Dune__exe__Main.main in file "backend/cn/main.ml", line 222, characters 8-73
    Called from Cmdliner_term.app.(fun) in file "cmdliner_term.ml", line 24, characters 19-24
    Called from Cmdliner_eval.run_parser in file "cmdliner_eval.ml", line 34, characters 37-44
```

[...]

PR #251, opened 2024-05-08
SMTLIB backend for CN

yav:
This PR implements a new CN solver back-end that communicates with SMT solvers as a separate process via smt-lib.
This makes it easier to experiment with different solvers and, hopefully, does not have too big of an impact on performance.


cp526 (approved):
This looks really good to me, and much cleaner than our current setup in various ways!


Miscellaneous comments below.

- Regarding `num_adds`: IIUC all "num_" functions work both for Int and Real. I don't know if the number literal "0" would work for Real (not that we use reals currently)

- Regarding avoiding name collisions. Where we have a value of type Sym.t, that's basically a number plus some pretty-print string. These symbols are globally unique, so to get out a unique string one could do `Sym.pp_string x ^ "_" ^ string_of_int (Sym.num x)` or similar. For struct, record, or constructor field names of type Id.t, we might make those unique by prefixing their names with the struct/record/datatype they're from.

- I see `defined_lets` in solver2.ml. If we used SMTLIB let-bindings, would those still be needed?

- Regarding "CN_CType": this one isn't so clear to me at the moment. We want to eventually change the Cerberus/CN pipeline to get rid of these altogether. For the moment I think C-types could be an uninterpreted sort in SMT, where for each value of C-type translated to SMT we could introduce a fresh constant, keeping a map so any C-type value that has already been translated in the past is mapped to the same constant as before (so the equalities work out as expected).

- Regarding "sort_record": CN record types should, by this time, already be normalised so their fields are sorted by field name. Maybe we could have an is_sorted assertion rather than sort_record to catch cases where they're not. (Also relevant to line 307)

- Regarding `translate_base_type`. 

   - Are you planning to map signed and unsigned bitvectors to the same "raw" bitvector sort in SMT? If that works (including the SMT-to-CN translation) that's very nice. I think this came up in our chat back then.

   - `Alloc_id`. CN has a `use_vip` switch. When that's off CN's current solver.ml translation maps `Alloc_id` to type Unit (so that no two different alloc IDs can exist), effectively turning all VIP-specific extra checks in CN off. If possible, once it's all working, it would be nice to replicate that.

- Regarding the CN-to-SMT term translation.

   - `MapDef`. IIRC, CN's current translation of index term MapDef to SMT should be dead code; the only purpose of MapDef (= lambda) in CN's current state should as an SMT-to-CN translation target when asking the solver for a model of an array value.

   - `Apply`. If the `Def` case *doesn't* apply this declares an uninterpreted function. 

   - `Let`. SMTLIB has let-bindings (unlike the Z3 OCaml API); maybe it would make sense to use those to simplify our translation of let-binders (and perhaps get better performance by letting the solver decide when to inline them)?

   - `Match`. Apparently SMTLIB also has `match`, which could save us from implementing it manually.

- Regarding `declare_datatype`. This probably doesn't do mutually recursive types yet, I guess. For handling those, `global.datatype_order` is useful. It's the topological order of datatypes, a list of groups of mutually recursive datatypes (ordered so later groups can depend on earlier groups). 

- Regarding `declare_solver_basics`. Probably the `s.globals.struct_decls` have to be declared before the `s.globals.datatypes`. The latter are CN datatypes, which can depend on C struct declarations (but not the other way around).

septract:
It's so cool that this now passes the test suite! 🎉

yav:
@septract This is still using the current solver (tbh, I am not sure why the tests weren't passing for a while).  To try the new solver one needs to run the `swap_solvers` script then rebuild.   I was thinking of swapping to the new solvers because the tests pass for me locally, but we'll  need to make sure that the `cvc5` or `z3` executable is installed in the CI environment, and I haven't looked at how to do that yet

cp526:
That looks really good! A couple of comments for the parts that I've managed to go over

# simple_smt.ml:

584: does this not need to take lambdas+quantifiers into account?


# solver.ml:

32, 36: I think we might have to prefix those with the struct/datatype tag (of type Sym.t): AIUI selector names have to be unique across sorts in smtlib.

258: maybe good to rename `base` and `offset` to `alloc_id` and `addr`, respectively.

305: I think for CN-to-SMT we do need to have some mapping for C-types. (The other direction, SMT-to-CN is nice-to-have but not so important for C-types, I think.) Could we handle C-types-values as uninterpreted constants/nullary functions, of an uninterpreted sort 'C-type' (along the same lines as the current solver.ml)?

septract:
Datapoint: this branch builds and runs just fine for me. 

Some of the CI tests fail for timeout reasons. I think we should decide a reasonable way to fix this before merging.

kmemarian:
Please remove the merge commits and instead rebase before this can be merged to master

yav:
@cp526 OK, I think I've made the changes you suggested.  The one difference is I am compiling C types to unique `int`s, instead of uninterpreted constants.  This has the benefit that the solver should be able to differentiate between different C types, and also we can parse them back into counter examples, because we know which `int` corresponds to which type.
What do you think?

@kmemarian OK, I'll do a squash and rebase when we are ready to merge.  It might have been nice to preserve the history, but just rebasing on top of master is quite a bit of work, as I'd have to resolve conflicts all over again.

kmemarian:
@yav ok, thanks

cp526:
@yav This looks great, thanks very much!

Just to clarify the part about C-types: these (I think) don't come up in the existing CI tests; it will be tests involving function pointers where C-types-as-values make it into the SMT problem.

I think `struct_field_name` should have `tag: Sym.t` as an extra argument and include `tag` in the generated name: then if two different struct types both involve the same field name the generated SMT names will be disjoint; and same thing for `datatype_field_name`, which should probably take the datatype tag as an argument to make the datatype field names globally unique. 

But this shouldn't delay the pull request. 

Looking at the failing CI tests: 

- `broken/error-crash/00005_bit_switch.c` times out, but is expected to crash. In principle it's possible that that tests exposes some subtle change in the behaviour between old and new smt backend. But it seems much more likely that the new SMT backend just times out before CN reaches that crash. Since that crash is already reported in `https://github.com/rems-project/cerberus/issues/233`, I think that's fine, and we just have to look into fixing any timeouts anyway.
 
- `broken/error-timeout/00011_dependen_specifications.c` used to time out but doesn't any longer. Again, that looks fine, we can re-categorise that test.

@kmemarian Looks all good, we can do the squash-and-merge/

@yav Thanks very much again!

kmemarian (approved):

PR #336, opened 2024-06-19
CN: First round of mli files, plus easy tidying

bcpierce00:
CN: Add mli files for a bunch of modules, plus some easy tidying

### DIFF
--- a/backend/cn/false.mli
+++ b/backend/cn/false.mli
@@ -1,0 +1,10 @@
+(* Module False - TODO: BCP: What is it?? *)
+
+(* TODO: BCP: Should this be abstract? It's only used concretely in one place... *)
+type t = False
+
+val subst : 'a -> t -> t
+val pp : t -> Pp.document
+
+module SymSet : Set.S with type elt = Sym.t
+val free_vars : t -> SymSet.t
--- a/backend/cn/id.mli
+++ b/backend/cn/id.mli
@@ -1,0 +1,19 @@
+(* Module Id -- Utility functions for Cerberus identifiers
+
+   This module adds a number of useful functions on identifiers to the
+   ones already provided by Cerberus. *)
+
+(* TODO: BCP: I'm a bit surprised that some of these are not already provided by Cerberus! *)
+
+type t = Cerb_frontend.Symbol.identifier
+
+val s : t -> string
+val loc : t -> Cerb_location.t
+val pp_string : t -> string
+val pp : t -> PPrint.document
+val equal : t -> t -> bool
+val compare : t -> t -> int
+val parse : Cerb_location.t -> string -> t
+val id : string -> t
+val is_str : String.t -> t -> bool
+val subst : 'a -> 'b -> 'b
--- a/backend/cn/list.ml
+++ b/backend/cn/list.ml
@@ -1,8 +1,11 @@
+(* This file re-exports the standard List library functions together
+   with a bunch of other useful list utilities. *)
+(* TODO: BCP: Probably not worth bothering with an mli file for this one... *)
+
 include Stdlib.List
 
 let concat_map (f : 'a -> 'b list) (xs : 'a list) : 'b list =
     concat (map f xs)
-
 
 let rec filter_map (f : 'a -> 'b option) (xs : 'a list) : 'b list =
   match xs with
@@ -12,13 +15,11 @@ let rec filter_map (f : 'a -> 'b option) (xs : 'a list) : 'b list =
      | None -> filter_map f xs
      | Some y -> y :: filter_map f xs
 
-
 let rec equal (equality : 'a -> 'a -> bool) (xs : 'a list) (xs' : 'a list) : bool =
   match xs, xs' with
   | [], [] -> true
   | x::xs, x'::xs' -> equality x x' && equal equality xs xs'
   | _, _ -> false
-
 
 let rec compare (comparison : 'a -> 'a -> int) (xs : 'a list) (xs' : 'a list) =
   match xs, xs' with
@@ -29,14 +30,12 @@ let rec compare (comparison : 'a -> 'a -> int) (xs : 'a list) (xs' : 'a list) =
   | [], _ -> -1
   | _, [] -> 1
 
-
 let mem equality y xs =
   let rec aux = function
     | [] -> false
     | x :: xs -> equality y x || aux xs
   in
   aux xs
-
 
 let assoc_opt (equality : 'k -> 'k -> bool) (k : 'k) (l : ('k * 'v) list) : 'v option =
   match find_opt (fun (k',_) -> equality k k') l with
@@ -51,13 +50,8 @@ let mem_assoc (equality : 'k -> 'k -> bool) (k : 'k) (l : ('k * 'v) list) : bool
   | Some _ -> true
   | None -> false
 
-
 let json jsonf list =
   `List (map jsonf list)
-
-
-
-
 
 let map_split (f : 'a -> 'b * 'c) (xs : 'a list) : 'b list * 'c list =
   fold_right (fun x (ys, zs) ->
@@ -65,21 +59,18 @@ let map_split (f : 'a -> 'b * 'c) (xs : 'a list) : 'b list * 'c list =
       (y :: ys, z :: zs)
     ) xs ([], [])
 
-
-
 let map_fst (f : 'a -> 'c) (xs : ('a * 'b) list) : ('c * 'b) list =
   map (fun (a, b) -> (f a, b)) xs
 
 let map_snd (f : 'b -> 'c) (xs : ('a * 'b) list) : ('a * 'c) list =
   map (fun (a, b) -> (a, f b)) xs
 
-
 let rec separate_and_group (by : 'b -> 'a option) : 'b list -> ('a option * 'b list) list =
   function
   | [] -> []
-  | b :: bs -> 
+  | b :: bs ->
     match by b, separate_and_group by bs with
-    | Some a, ((None, bs') :: abs) -> 
+    | Some a, ((None, bs') :: abs) ->
         (Some a, bs') :: abs
     | Some a, abs ->
         (Some a, []) :: abs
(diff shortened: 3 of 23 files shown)


bcpierce00:
This is my first whack at creating mli files for a number of the CN modules.  It's a large PR, but I hope easy to review, since there are no functionality changes at all.  

Creating the mli files did identify some dead code and a couple of opportunities for easy renaming to remove potentially confusing duplicate names for types.  

There's still some work to do on documenting the new mli files -- look for `TODO: BCP` if you want to see where.  But I think this can be merged now if CN folks are OK with that.

cp526:
Thanks very much!

dc-mak:
Can we please take a little longer to discuss some of the issues highlighted by this PR? E.g. some of the TODOs and how to handle them going forward?

backend/cn/false.mli
dc-mak, line 1: This module is functioning as a unit type to end specifications. I think `False` is a misleading name and it should really just be deleted and replaced wherever with a unit type.

backend/cn/id.mli
```diff
@@ -0,0 +1,19 @@
+(* Module Id -- Utility functions for Cerberus identifiers
+
+   This module adds a number of useful functions on identifiers to the
+   ones already provided by Cerberus. *)
+
+(* TODO: BCP: I'm a bit surprised that some of these are not already provided by Cerberus! *)
```
dc-mak, line 6: I think the main purpose of this module is shortening some names, but I think it goes a little too far...

backend/cn/id.mli
```diff
@@ -0,0 +1,19 @@
...
+
+type t = Cerb_frontend.Symbol.identifier
+
+val s : t -> string
```
dc-mak, line 10: E.g. this should be `to_string` by convention

backend/cn/id.mli
```diff
@@ -0,0 +1,19 @@
...
+val loc : t -> Cerb_location.t
+val pp_string : t -> string
```
dc-mak, line 12: And this should be deleted since it aliases the to_string...

backend/cn/id.mli
```diff
@@ -0,0 +1,19 @@
...
+val pp : t -> PPrint.document
+val equal : t -> t -> bool
+val compare : t -> t -> int
+val parse : Cerb_location.t -> string -> t
```
dc-mak, line 16: This should be `make` or (whatever the convention is)

backend/cn/id.mli
```diff
@@ -0,0 +1,19 @@
...
+val id : string -> t
```
dc-mak, line 17: This should probably be deleted

bcpierce00:
Of course.  I suggested merging before discussing only to make sure that
the PR doesn't get even bigger in the meantime.  But it would certainly be
good to use the opportunity to make decisions about things.

Where shall we discuss?

backend/cn/id.mli
```diff
@@ -0,0 +1,19 @@
...
+val is_str : String.t -> t -> bool
```
dc-mak, line 18: This should probably be renamed to `equal_string` and have the `t` first by convention.

backend/cn/id.mli
```diff
@@ -0,0 +1,19 @@
...
+val subst : 'a -> 'b -> 'b
```
dc-mak, line 19: I'm confused by this

backend/cn/list.ml
```diff
@@ -1,9 +1,12 @@
+(* This file re-exports the standard List library functions together
+   with a bunch of other useful list utilities. *)
+(* TODO: BCP: Probably not worth bothering with an mli file for this one... *)
```
dc-mak, line 3: Correct, I have a plan to replace some of the utils with the Base library which has more readable and sensible conventions IMO

backend/cn/locations.mli
```diff
@@ -1,14 +1,25 @@
-(* taking things from ocaml_locations *)
+(* Module Locations -- Utility functions for Cerberus locations
+
+   This module adds a number of useful functions on locations to the
+   ones already provided by Cerberus. *)
 
+(* taking things from ocaml_locations *)
+(* TODO: BCP: What does that mean? *)
```
dc-mak, line 7: Means "copying". There are a few of these littered around the code base and sometimes they are helpful (e.g. if pointing to file or function) and sometimes they are obtuse).

backend/cn/locations.mli
```diff
@@ -1,14 +1,25 @@
...
 
+(* TODO: BCP: I think this type could be made abstract with only a little
+   work -- its identity seems to be used in just a few places in other
+   modules. *)
 type t = Cerb_location.t
 
+(* TODO: BCP: It's a bit hideous to to export two different names for the same (morally abstract) type. *)
```
dc-mak, line 14: Agreed

backend/cn/locations.mli
```diff
@@ -1,14 +1,25 @@
...
+   This module adds a number of useful functions on locations to the
+   ones already provided by Cerberus. *)
 
+(* taking things from ocaml_locations *)
+(* TODO: BCP: What does that mean? *)
 
+(* TODO: BCP: I think this type could be made abstract with only a little
+   work -- its identity seems to be used in just a few places in other
+   modules. *)
 type t = Cerb_location.t
```
dc-mak, line 12: I'm not sure if that'd be worth it? But I'm open to it.

backend/cn/locations.mli
```diff
@@ -29,6 +40,7 @@ val head_pos_of_location : t -> string * string
 
 val unpack : t -> Cerb_location.t
 
+(* TODO: BCP: What's to do? *)
 (* todo *)
```
dc-mak, line 44: No idea. Probably just delete.

backend/cn/pp.mli
```diff
@@ -0,0 +1,177 @@
+(* Module Pp -- Pretty printing *)
+
+(* TODO: BCP: This file could still use some thinning and reorganization! *)
```
dc-mak, line 3: I suspect one way to proceed might be to have `pp.ml` be just `include PPrint` (or delete it and get users to rename it wherever it is used). Everything else that's currently in `pp.ml` could be turned into `pp_utils.ml`

backend/cn/symRel.mli
```diff
@@ -0,0 +1,89 @@
+(* Module SymPair -- TODO: BCP: What is it for?? *)
+
+(* TODO: BCP: Some filenames use_underscores, while some areInCamlCase
+   -- any compelling reason for that? *)
```
dc-mak, line 4: No. We need to pick a convention and just go for it. My vote is snake_case (easier on the eyes).

backend/cn/symRel.mli
```diff
@@ -0,0 +1,89 @@
+(* Module SymPair -- TODO: BCP: What is it for?? *)
```
dc-mak, line 1: Nothing. It may be deleted (including in `explain.ml`).

backend/cn/symRel.mli
dc-mak, line 1: This file and its implementation can be deleted.

dc-mak (commented):
Apologies for the many comments, but hopefully we can start putting in some deprecation warnings into the code base for things which should be renamed and suggest alternatives.

cp526:
Regarding False. ArgumentTypes.t is polymorphic in what the return type is: e.g. ArgumentTypes.t with ReturnTypes.t is the OCaml type of CN function specifications. 

False is the instance we use for the type of Non-return procedures (used in Core for implementing C goto labels and loop control-flow). Hence the name `False`: following goto, one gets to assume logical false at the call site.

So this one I want to keep, not replace with `unit`.

dc-mak:
Is there a reason we can't use `Constraint ((IT.bool_ false (Locations.other __FUNCTION__), _, I ())` ?

backend/cn/false.mli
bcpierce00: I will move this comment to the .mli file, marked TODO: DCM:

bcpierce00: (Actually, on second thought I will leave it to Dhruv or Christopher to do that, for this and other comments that I can't fix myself, unless someone tells me it's useful.)

backend/cn/list.ml
bcpierce00: I'll delete my comment

backend/cn/locations.mli
bcpierce00: I'll delete my comment.

backend/cn/locations.mli
bcpierce00: I'll try to unify.

backend/cn/locations.mli
bcpierce00: I'll see if it's actually easy.

backend/cn/locations.mli
bcpierce00: OK

backend/cn/pp.mli
bcpierce00: I'll just add your comment to the file, since I dojn't think we're going to do it now.

backend/cn/symRel.mli
bcpierce00: Agreed.

backend/cn/symRel.mli
bcpierce00: Will do.

bcpierce00:
What did you mean by "hopefully we can start putting in some deprecation warnings into the code base for things which should be renamed and suggest alternatives"?  What things did you have in mind?

dc-mak:
I put the comments inline on the files themselves; you should be able to see them if you scroll up?
*`Id.s` should IMO be `to_string` 
* `Id.pp_string` should be deleted or deprecated
* `Id.parse` would be clearer as `Id.make` 
* `Id.id` should really be deprecated/deleted since the location info is not useful unless used at the call site.
* `Id.is_str` can maybe stay as is or be renamed to `Id.equal_string` 

These don't have to be done immediately but even TODO comments would I think help log the direction of travel.

bcpierce00:
But why not just go ahead and delete or rename them?  Do we not control all
the code that depends on this module?

dc-mak:
Sure, I was thinking that if there were like 500 uses of all those things it might be a bit tedious to replace them all (but I realise now that in this instance the type isn’t changing so a find and replace would work).

I have a separate branch trying to use the Base.List module instead of our increasingly large hand-rolled one and there are almost a 1000 uses there to change (adding labels to arguments which didn’t have them before). That’s the context in which I was considering effort vs reward, and why depreciations may be a gentler transition even if we own the whole codebase.

PR #341, opened 2024-06-22
[CN-Test-Gen] Add CLI flag for test generation

ZippeyKeys12:
I'd like to avoid conflicts every time `main.ml` changes, so this PR just adds the code that will call test generation.

### DIFF
--- a/backend/cn/main.ml
+++ b/backend/cn/main.ml
@@ -125,12 +125,20 @@ let check_input_file filename =
     if not (ext ".c" || ext ".h") then
       CF.Pp_errors.fatal ("file \""^filename^"\" has wrong file extension")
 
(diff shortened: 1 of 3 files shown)


dc-mak (commented):
Hi Zain, appreciate the commit but I worry the logic is getting a little too complicated.

Ideally we would shift to using [sub commands](https://github.com/rems-project/cerberus/issues/343) very soon. 

Perhaps a little frustrating, but my vote is to hold fire whilst we discuss this.

rbanerjee20:
Hi Zain, could you run `git pull --rebase master` on this branch? I can test out the runtime checking tool on this branch after you've done that. I'll let you know if it is behaving as expected and then we can merge this in.

Issue #337, opened 2024-06-20
[CN] Allow `'` in numeric constants


yav:
It'd be nice to allow `_` in numeric constants, so you can write things like `-2147483648_i32` or `0x_FFFF_FFFF`.

If we do this, it would also be nice to update the pretty printers to add `_` before the type suffix of constants (e.g., `0_i32` instead of `0i32`).

bcpierce00:
Excellent onboarding project.

dc-mak:
Unfortunately, to match C23 and C++, the separator must be `'`.

septract:
Addressed in #379, but I used `_` as originally proposed 

@dc-mak why would this have to be `'`? The syntax only appears inside magic comments, which are not specified by C23 / C++, surely?

dc-mak:
For consistency between the spec and C and code re-use in the lexer (e.g. also handling hex, binary and octal). This feature is being handled in the ongoing C/CN parser split.

dc-mak:
This will be shipped as part of: https://github.com/rems-project/cerberus/pull/342
However, if the feature is urgent we can consider separate PRs.
(Sorry @septract for not updating the project board and the time spent).

yav:
@dc-mak having to write `0i32` is already different from C though, is that changing?

thatplguy:
FWIW I agree with @dc-mak – we should try to follow C conventions and standards where possible, even in magic comments, unless there's a reason not to.  Thanks @dc-mak for adding this in the parser refactor.

dc-mak:
`0i32` is a limitation of the term language type system and can be overcome with a bit of engineering. It was more a pragmatic work-around to keep moving when we transitioned from integers to bit vectors rather than a design choice.

PR #342, opened 2024-06-22
Separating the C and CN parsing

kmemarian:
This creates a separate lexer and parser for CN.

There are a few questions for @cp526 and @dc-mak marked with `TODO(Christopher/Dhruv)` in the code.

And there is a hack for dealing with the parsing of C type-names discussed in the commit message of c96a1adcf7c743f8150c8f1669e93933fea2249f.

This is still work-in-progress, the CN parser doesn't produce pretty messages so some CI and tutorial tests fail.


cp526:
Thanks very much! 

AFAICS `CN_bool` and `cn_function` can go away. Not clear to me at the moment whether `boolean` might have a use if there's a clash of names between C (integer) booleans and CN (proper) booleans.

cp526:
Regarding the question of parsing C-types within CN: maybe the lexer hack is less problematic than it seems: there's a small set of language constructs only that can take C-types (`Owned`, `array_shift`, and maybe nothing else?), and those all require parentheses following the `<ctype>` part for the remaining arguments. I wonder whether that makes the syntax unambiguous wrt the C-type angle-bracket notation so there's never a need to re-lex?

parsers/c/cn_lexer.mll
```diff
@@ -0,0 +1,261 @@
...
+  "accesses", ACCESSES;
+  "alloc_id", ALLOC_ID;
+  "apply", APPLY;
+  "array_shift", ARRAY_SHIFT;
+  "assert", ASSERT;
+  "Block", BLOCK;
+  "bool", BOOL; (* shared with C23 *)
+  "boolean", BOOL; (* TODO(Christopher/Dhruv): are all these variants of BOOL needed? *)
+  "CN_bool", BOOL; (* TODO(Christopher/Dhruv): are all these variants of BOOL needed? *)
+  "cn_function", FUNCTION; (* TODO(Christopher/Dhruv): is this variant still needed? *)
```
dc-mak, line 18: Delete them all!

parsers/c/cn_lexer.mll
```diff
@@ -0,0 +1,261 @@
...
+let digit = ['0'-'9']
+let hexadecimal_digit = ['0'-'9' 'A'-'F' 'a'-'f']
+let octal_digit = ['0'-'7']
+let nonzero_digit = ['1'-'9']
+
+let hexadecimal_prefix = "0x" | "0X"
+let hexadecimal_constant = hexadecimal_prefix hexadecimal_digit+
+
+(* C23 binary constant, but omitting ' separators for now *)
+(* TODO(Christopher/Dhruv): add support for ' separators to be in line with C23 ? *)
```
dc-mak, line 103: Yes please! https://github.com/rems-project/cerberus/issues/337

parsers/c/cn_lexer.mll
```diff
@@ -0,0 +1,261 @@
...
+let integer_constant =
+    decimal_constant
+  | octal_constant
+  | hexadecimal_constant
+  | binary_constant
+
+(* Identifiers - we follow C11 *)
+(* TODO(Christopher/Dhruv): do you care about following C11 closely for this
+    (e.g. with respect to the universal character stuff), or should we simplify
+    like we already do for uppercase names. *)
```
dc-mak, line 122: I vote simplify - we can always add it later if requested.

parsers/c/cn_lexer.mll
```diff
@@ -0,0 +1,261 @@
...
+  | '<' ([^'='][^'<' '>']+ as str) '>'
+      {
+        try
+          LT_CTYPE_GT (C_parser.type_name_eof C_lexer.lexer (Lexing.from_string str))
+        with
+          | _ ->
+              relexbuf_opt := Some (Lexing.from_string (str ^ ">"));
+              LT
+      }
+
```
dc-mak, line 177: This is an amazing hack. But I do worry it might be a bit too hacky. Can it be more conservative? Specifically, I wanted to have `<-` available as a token. We do need to discuss this properly though @cp526

dc-mak (commented):
Lgtm, though this ctype issue is gonna be annoying.

parsers/c/cn_lexer.mll
```diff
@@ -0,0 +1,261 @@
...
+let keywords: (string * Cn_tokens.token) list = [
+  "accesses", ACCESSES;
+  "alloc_id", ALLOC_ID;
+  "apply", APPLY;
+  "array_shift", ARRAY_SHIFT;
+  "assert", ASSERT;
+  "Block", BLOCK;
+  "bool", BOOL; (* shared with C23 *)
+  "boolean", BOOL; (* TODO(Christopher/Dhruv): are all these variants of BOOL needed? *)
+  "CN_bool", BOOL; (* TODO(Christopher/Dhruv): are all these variants of BOOL needed? *)
```
dc-mak, line 17: These two can go.

dc-mak:
> Thanks very much!
> 
> AFAICS `CN_bool` and `cn_function` can go away. Not clear to me at the moment whether `boolean` might have a use if there's a clash of names between C (integer) booleans and CN (proper) booleans.

We're still using `cn_function` for constants! https://rems-project.github.io/cn-tutorial/#_constants

cp526:
> > Thanks very much!
> > AFAICS `CN_bool` and `cn_function` can go away. Not clear to me at the moment whether `boolean` might have a use if there's a clash of names between C (integer) booleans and CN (proper) booleans.
> 
> We're still using `cn_function` for constants! https://rems-project.github.io/cn-tutorial/#_constants

Yes, that's true.

parsers/c/cn_lexer.mll
cp526: The idea of `...<...>` was to have a C++-like syntax for type parameters, which would be good to keep if we can. What's the use of `<-`?

parsers/c/cn_lexer.mll
dc-mak, line 18: This stays.

parsers/c/cn_lexer.mll
yav: I think types always start with a letter (or maybe `_`), so you could probably change it to only try for a type, if the `<` is followed by one of these. I think should allow you to have `<-` later if you wanted it.

Otoh, I don't think this would work if you ever wanted types that have instantiations in them, but we probably don't have those at the moment.

PR #344, opened 2024-06-24
CN + Exec: Run spec checks before decoration

dc-mak:
Commit a18c9da798d6d6ea7b89813c7fa1faef050e0fc0 implemented this with a bug: the proof was running even after decoration. Hence, it was reverted in 2cbe6f3b989ddb4cc048f5ed39f79710dba437d8.

This commit should implement the behaviour correctly now. However, for a more robust solution, we should look into refactoring using subcommands.

### DIFF
--- a/backend/cn/main.ml
+++ b/backend/cn/main.ml
@@ -125,13 +125,6 @@ let check_input_file filename =
     if not (ext ".c" || ext ".h") then
       CF.Pp_errors.fatal ("file \""^filename^"\" has wrong file extension")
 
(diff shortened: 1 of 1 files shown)


dc-mak:
This PR should only be approved and merged after confirmation that executable spec tests are passing. (Though it would be surprising if they weren't, and would indicate the tests are broken and need fixing, now that the change itself is not buggy.)

rbanerjee20 (approved):
LGTM, have checked out this branch and executable spec tests are passing+failing as on master.

Issue #343, opened 2024-06-24
[CN] Use subcommands `cn verify`, `cn reify`, `cn lemmas` etc


dc-mak:
Care needs to be taken so that CI and other scripts are updated simultaneously.
https://github.com/mjambon/cmdliner-cheatsheet?tab=readme-ov-file#subcommands

ZippeyKeys12:
@dc-mak  Shouldn't we make successor issues for pulling out the executable spec and lemma export into their own subcommands as well?

dc-mak:
No-one's working on lemmas right now and so that can be postponed. I suppose we could for executable spec though.

PR #346, opened 2024-06-24
CN: Group comamnd line options

dc-mak:
This commit groups the current CN options to make it clearer which are used for what.

This could also be useful in preparation for switching to using [subcommands](https://github.com/rems-project/cerberus/issues/343).

### DIFF
--- a/backend/cn/main.ml
+++ b/backend/cn/main.ml
@@ -239,169 +239,10 @@ open Cmdliner
 
 
 (* some of these stolen from backend/driver *)
+module Common_flags = struct
 let file =
   let doc = "Source C file" in
   Arg.(required & pos ~rev:true 0 (some string) None & info [] ~docv:"FILE" ~doc)
-
-let incl_dirs =
-  let doc = "Add the specified directory to the search path for the\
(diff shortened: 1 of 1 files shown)


dc-mak:
@cp526 could you check over this? Not familiar with all of these options.

dc-mak:
And @rbanerjee20 too.

rbanerjee20:
For the CN executable spec flags, the grouping of those 3 flags is correct, but I would prefer if the module was named something else, and possibly with a `flag` suffix since without this it might be confused with existing modules like `Executable_spec`. E.g. `Runtime_checking_flag`, `Executable_spec_flag` (the latter would match the existing naming convention for CN executable spec modules).

cp526:
Agreed about the `flag` suffix in the module names.

The flag `use_peval` is probably specific to CN proof; I don't think the executable-spec tool can produce json (though maybe that would be useful in the future?), so the `json` flag is also specific to CN proof.

dc-mak:
Thanks - addressed. Any other comments or good to merge?

PR #347, opened 2024-06-24
CN: Use Base.List library instead of hand-rolling

dc-mak:
This PR demonstrates that we can migrate over to using the [Base](https://opensource.janestreet.com/base/) if we wanted to, and bring with it the benefits of a safer, more full-featured and more consistent library. The biggest change would be to `List`, though I've found myself wanting many `Option` functions (e.g. `value_map`) too.

It took less than an hour (though I did have a false start last Friday) to get to this point and would be finished in about another hour's work; the following functions remain:
```
15:01 ➜  cerberus git:(cn-base-list) grep 'List\.Old\.\w\+' -R backend/cn -o | awk -F':' '{ print $2; }' | sort | uniq -c | sort
      1 List.Old.compare
      1 List.Old.find
      1 List.Old.find_map
      1 List.Old.fold_left_map
      1 List.Old.iter2
      1 List.Old.last
      1 List.Old.map_split
      1 List.Old.nth_list
      1 List.Old.partition_map
      2 List.Old.is_empty
      2 List.Old.iteri
      2 List.Old.sorted_and_unique
      3 List.Old.sort_uniq
      4 List.Old.add
      4 List.Old.equal
      6 List.Old.init
      7 List.Old.map_snd
      7 List.Old.mem
      9 List.Old.sort
     13 List.Old.assoc
     13 List.Old.assoc_opt
     26 List.Old.fold_right
     40 List.Old.fold_left
```

### DIFF
--- a/backend/cn/report.ml
+++ b/backend/cn/report.ml
@@ -649,7 +649,7 @@ let read_file filename =
     | _ -> None
 
 let make filename source_filename_opt (report: report) =
-  let n_pages = List.Old.length report.trace in
+  let n_pages = List.length report.trace in
   assert (n_pages > 0);
 
   let _menu = div ~id:"menu"
--- a/backend/cn/report.ml
+++ b/backend/cn/report.ml
@@ -64,10 +64,10 @@ let div ?clss ?id elements =
 (* let pre code = enclose "pre" code *)
 (* let body elements = enclose "body" (list elements) *)
 let h i title body = list [enclose ("h"^string_of_int i) title; body]
-let table_row cols = enclose "tr" (list (List.Old.map (enclose "td") cols))
-let table_head_row cols = enclose "tr" (list (List.Old.map (enclose "th") cols))
+let table_row cols = enclose "tr" (list (List.map ~f:(enclose "td") cols))
+let table_head_row cols = enclose "tr" (list (List.map ~f:(enclose "th") cols))
 let table_head cols = enclose "thead" (table_head_row cols)
-let table_body rows = enclose "tbody" (list (List.Old.map table_row rows))
+let table_body rows = enclose "tbody" (list (List.map ~f:table_row rows))
 let table head rows = enclose "table" (list [table_head head; table_body rows])
 let table_without_head rows = enclose "table" (list [table_body rows])
 let details summary more = enclose "details" (list [enclose "summary" summary; more])
@@ -122,7 +122,7 @@ let make_predicate_hints predicate_hints =
   lguard predicate_hints (fun predicate_hints ->
     h 1 "Possibly relevant predicate clauses" (
       table ["condition"; "clause"]
-        (List.Old.map (fun pce -> [Pp.plain pce.cond; Pp.plain pce.clause]) predicate_hints)
+        (List.map ~f:(fun pce -> [Pp.plain pce.cond; Pp.plain pce.clause]) predicate_hints)
     )
   )
 
@@ -132,7 +132,7 @@ let make_predicate_hints predicate_hints =
 (*     | [] -> "(no available resources)" *)
 (*     | _ -> *)
 (*       table ["resource"; "byte span and match"] *)
-(*         (List.Old.map (fun re -> [Pp.plain re.res; Pp.plain re.res_span]) resources) *)
+(*         (List.map ~f:(fun re -> [Pp.plain re.res; Pp.plain re.res_span]) resources) *)
 (*   ) *)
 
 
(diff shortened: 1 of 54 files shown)


backend/cn/report.ml
```diff
@@ -467,7 +467,7 @@ function multiline_highlight(start_line, start_col, end_line, end_col) {
         cn_code.scrollTop = div.offsetTop
       }
       hl = document.createElement("span")
-      hl.classList.add("hl")
+      hl.classList.Old.add("hl")
```
dc-mak: Woops this is a mistake

backend/cn/report.ml
```diff
@@ -563,9 +563,9 @@ function create_line(n, str) {
   nb_div = document.createElement("div")
   str_div = document.createElement("div")
   nb_div.textContent = n
-  nb_div.classList.add("nb")
+  nb_div.classList.Old.add("nb")
```
dc-mak: Also mistake

backend/cn/report.ml
```diff
@@ -563,9 +563,9 @@ function create_line(n, str) {
...
   str_div.textContent = str
-  str_div.classList.add("line")
+  str_div.classList.Old.add("line")
```
dc-mak: Also mistake

dc-mak:
@cp526 Yay or nay? The rebase took 2.5x longer than the original work, so if it's a yay, I'd like to get it merged in asap. Can finish the rest piecemeal.

@rbanerjee20 Please can you confirm executable spec things as intended. All changes are mere renamings.

cp526:
Thanks for investigating this @dc-mak ! 

I'm not completely convinced it's a good idea though. Our list module is only 100 lines of code, including for some helper functions that Base.List won't have. Unless we can save lots of code elsewhere, I'm not sure moving to `Base` is worth the cost of this big additional dependency.

(Also, the rest of Cerberus won't be using Base; don't know if the inconsistency in standard libraries would create problems.)

dc-mak:
* We'll be splitting out of Cerberus at some point soonish, so that's a time limited consideration.
* Base is intentionally quite a small and _very_ common dependency (it's likely already installed on your machine). It's not the same as `Core` or `Core_kernel`, and _11x smaller_ than the Z3 dependency.
* It's not just saving 100 lines of code, it's to make it easier to onboard new developers. We've already had two instances where (a)  developer was surprised we override the `List.mem` function (b) developer accidentally used `==`. Both of these would be prevented by using `Base`.
* The number of developers familiar with `Base` and its naming conventions will always be higher than ones familiar with our hand-rolled ones (it's used in Real World OCaml).
* We won't have issues with the backporting stdlib functions (which has happened a couple of times already I think).
* If we ever want to cache any complex data (Sets, Maps, Hashtables etc.) Base will make this very straightforward.
* It will make code easier to format `List.map xs ~f:(fun x -> multi-line function)`

It is highly likely that as the project grows, our in-built libraries will only grow, and at some point will want to switch over - so we should do it asap (like with the source location on index terms).

Anyways, this my final word of caution on the matter; you might wish to consult others and see what they say.

thatplguy:
FWIW I'm very much in favor of using standard libraries whenever possible.  As @dc-mak points out, it smooths onboarding for new developers, and it pays off down the road every time you use another library function rather than implementing your own.  Standard libraries are often better tested as well.

I'm not super familiar with the OCaml ecosystem, but depending on libraries used in Real World OCaml seems reasonable (?).

dc-mak:
## Background
Update: according to @cp526, when the [rmem](https://github.com/rems-project/rmem ) project used [OCamlBatteries standard library](https://ocaml-batteries-team.github.io/batteries-included/hdoc2/), it got stuck for some time on an old OCaml version because the dependency was not updated,

## Concerns 
1. Base may be become a sticking point for upgrading OCaml versions.
2.  And additional dependency, and its size.
3. Potentially confusing migration path (of temporarily having both, but incrementally deprecating the old).

To be clear, these were not objections, but neither were they approvals. 

Since the last update, there have been two large refactors:
1. Formatting the code: https://github.com/rems-project/cerberus/pull/403
2. CN as a library: https://github.com/rems-project/cerberus/pull/429

Thus this PR is out of date. Any future refactors may want to look into using OCaml ppx-rewriters (or perhaps good refactoring tools for OCaml will be available by then).

bcpierce00:
IIRC, the main argument for using Base.List was familiarity (and hence
smoothness of onboarding new devs).  This concern can be addressed by
making a local copy of Base.List (either wholesale or function-by-function)
instead of introducing a dependency.

dc-mak:
Perhaps, then confusion might arise as to why Option, Maps, Sets, etc. all have different style of APIs (Base makes uses of named arguments and conventions like `Module.t` as first argument, `*_exn` instead of  `_opt` etc).

bcpierce00:
Won't this be exactly the same confusion whether we use Base.List or a local copy?

dc-mak:
The intention was to move towards using Base overall (though the PR was specific to just Base.List to start with).

PR #364, opened 2024-07-04
Add a --version flag generated from git hash

### LINKED ISSUE #62: Add versioning to CN
Versioning is very helpful in diagnosing bugs with multiple users.
* Users can confirm they're using the latest, most up-to-date version
* Developers can ask users for the version they're using to be able to reproduce the bug they've encountered

I'd propose using [semantic versions](https://semver.org/), e.g. `vX.Y.Z`, where
* incrementing X indicates a non-backwards compatible change, typically a major release
* incrementing Y indicates new features
* incrementing Z indicates small improvements or bugfixes

Can we set up a release process for CN that supports
1. a `--version` flag that prints the current version (see schema above)
2. a GitHub version tag for version increments

More concretely, perhaps something like
* a CURRENT_VERSION file in the repo that contains the current version (e.g. `v0.0.1`)
* add `--version` flag that prints CURRENT_VERSION.  if the repo is dirty or the latest commit is not the same as the commit tagged with the current version, then add the git sha to the version, e.g. `v0.0.1-6aad28f0399643a55378c23a5dbbb0b45c243f27`)
* add a `make release` target (or shell script) that (a) checks that the repo is in a clean state, (b) pushes the CURRENT_VERSION as a GitHub tag

Then the release process is to update the CURRENT_VERSION via a PR, then run `make release` once the PR is merged.

Happy to hear alternatives though.

septract:
We should eventually add proper version numbers for CN, but as a stop-gap, it'd be useful to know the provenance of every `cn` binary. This PR builds a hash with every binary, and makes it available through the `--version` command line argument. (This idea proposed by @yav) 

How it works: 
* Generate a hash by calling `git rev-parse --short HEAD` in the top level Makefile
* Store the result in `backend/cn/version.ml`
* Modify the `cn` command-line options using the `Cmd.info ~version:XX` capability from `CmdLiner`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,15 +23,6 @@ else
     DUNEFLAGS=
 endif
 
-# Default hash in case git is not available
-DEFAULT_HASH := "unknown"
-GIT_HASH := $(shell if command -v git >/dev/null 2>&1 && git rev-parse --is-inside-work-tree >/dev/null 2>&1; then git rev-parse --short HEAD; else echo $(DEFAULT_HASH); fi)
-CN_VERSION_FILE := backend/cn/version.ml
-
-.PHONY: cn-version
-cn-version: 
-	echo "let git_hash = \"$(GIT_HASH)\"" > $(CN_VERSION_FILE)
-
 .PHONY: normal
 normal: cerberus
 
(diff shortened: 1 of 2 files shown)


septract:
@cp526 @dc-mak could one of you review this?

kmemarian:
@septract there is already an implementation of this as part of `cerb_frontend`. You should be able to access it using the `Cerb_frontend.Version.version` string from CN's main.

Though it is generated with a different git command `git describe --dirty --always` (see `tools/gen_version.ml`).

septract:
Even better. I don't feel strongly about what git command we use

PR #365, opened 2024-07-05
[CN] Make CN run both Z3 and CVC5 in CI, tidy up runner CI script

septract:
NOTE: depends on [this PR](https://github.com/rems-project/cn-tutorial/pull/37) in the `cn-tutorial` repo, which fixes a bug in the runner script handling of `cn` arguments. 

This PR also significantly tides up the runner script and shifts most of the logic over to the `cn-tutorial` MakeFile, and fixes a bug that was stopping some tests from running.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,7 +31,7 @@ jobs:
     
     - name: Restore cached opam
       id: cache-opam-restore
-      uses: actions/cache/restore@v3
+      uses: actions/cache/restore@v4
       with:
         path: ~/.opam
         key: ${{ matrix.version }}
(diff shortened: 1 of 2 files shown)


septract:
TODO: perhaps do the same for the cerberus-local test suite (run by the `tests/run-cn.sh` script) ?

septract:
CI now fails because z3 and cvc5 have different timing characteristics, so one example times out with cvc5. Annoying problem, unsure what to do about it.

bcpierce00:
Longer timeout?

cp526:
I've bumped up the timeout for now and merged this.

We would like the CI to find regressions in timing behaviour, though. So we'll probably want to extend our scripts to collect timing statistics and "publish" the results in a way that allows for easy comparison across commits + fail if the overall sum exceed some sensible limit.

septract:
@cp526 Agreed, I made an issue to capture this idea here: https://github.com/rems-project/cn-tutorial/issues/54

Issue #328, opened 2024-06-16
[CN] Implement `foo->bar` member syntax for pointers in spec language


dc-mak:

dc-mak:
Duplicate of https://github.com/rems-project/cerberus/issues/329

ZippeyKeys12:
There isn't a description, but wasn't this issue meant to be about `x->y` being sugar for `(*x).y`?

So not a duplicate of #329?

cp526:
Indeed

dc-mak:
Oops yes thanks

septract:
@dc-mak this issue can be closed

PR #368, opened 2024-07-06
[CN] Add parsing of a->b expressions

septract:
Work on addressing #328

Progress: 
- [x] Got CN to parse expressions by reusing the code from `CExpr_memberof` 
- [ ] Implement the semantics by desugaring

### DIFF
--- a/backend/cn/compile.ml
+++ b/backend/cn/compile.ml
@@ -199,6 +199,8 @@ let rec free_in_expr (CNExpr (_loc, expr_)) =
      free_in_exprs es
   | CNExpr_memberof (e, _id) ->
      free_in_expr e
+  | CNExpr_arrow (e, _id) ->
+     free_in_expr e
   | CNExpr_record members ->
      free_in_exprs (List.map snd members)
   | CNExpr_memberupdates (e, updates) ->
@@ -706,6 +708,9 @@ module EffectfulTranslation = struct
               IT (Nil (SBT.to_basetype item_bt), SBT.List item_bt, nil_loc) in
             return (List.fold_right cons es nil)
         | CNExpr_memberof (e, xs) ->
+           let@ e = self e in
+           translate_member_access loc env e xs
+        | CNExpr_arrow (e, xs) ->
            let@ e = self e in
            translate_member_access loc env e xs
         | CNExpr_record members ->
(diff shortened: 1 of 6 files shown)


septract:
I could use some input about how to complete this feature. My guess is this should work as follows: 
1. Add a constructor to the `cn_expr_` type, `CNExpr_arrow` (in `frontend/model/cn.lem`)
2. Parse expressions of the form a->b to `CNExpr_arrow` (in `parsec/c/c_parser.mly`)
3. Modify `desugar_cn_expr` to convert `CNExpr_arrow` to `CNExpr_memberof`, and inject a synthetic `CNExpr_deref` somewhere to complete the translation (in `frontend/model/cabs_to_ail.lem`). 

Alternative approach: I could also imagine changing the interpretation of `CNExpr_arrow` by modifying `translate_cn_expr` in `backend/cn/compile`. 

Is this on the right track @cp526 / @dc-mak ? 

(Note, we also have to do something similar for some a `CNExpr_arrowupdate` constructor)

dc-mak:
I think you might not even need to add the constructor - perhaps to get something working and test it, you can just have the parser `a->b` produce a `(member(deref(a), b)` in the Cabs and the rest of the pipeline should take care of it. Or you can do it at something like in `desugar_cn_expr`
```
arrow (a,b) ->
    desugar (member(deref(a), b))
```
Or in backend/cn/compile. 

I'm confused about an `arrowupdate` constructor though. If not too much of a tangent, can you explain?

septract:
I assumed that it would be needed as the equivalent of the `CNExpr_memberupdate` constructor, but perhaps I misunderstood what that constructor represents?

dc-mak:
The below passes CN (with arrow soon to be valid, but commented out for now). I can't see understand if arrow update makes sense as a concept.
```c
struct point {
    int x;
    int y;
};

void f() {
    struct point origin = { .x = 0, .y = 0 };
    /*@ assert (origin.x == 0i32); @*/ // -- member
    struct point not_origin = origin;
    not_origin.y = 1;
    /*@ assert (not_origin == { y: 1i32, ..origin});@*/ // member update
    struct point *p = &origin;
    // /*@ assert (p->x == 0i32); @*/ // -- arrow (when implemented)
    /*@ assert ((*p).x == 0i32); @*/ // -- arrow (as currently)
    struct point *q = &not_origin;
    /*@ assert (*q == { y: 1i32, ..*p});@*/ // how would arrow update be different/better?
}
```

septract:
Thanks, I think I just misunderstood the update construct! I'll get this PR fixed and submitted for review.

septract:
Closing in favor of #396

PR #381, opened 2024-07-11
CN-exec: Add support for leak-checking and free

dc-mak:
Adding support for frees doesn't make sense without also adding support for leak checking.

This commit adds both, and a few tests cases to check behaviour is as expected. Notably, it swaps the order of the post-condition checking and locals unmapping (if this were not done,
return_owned_local.error.runtime.c would erroneously pass runtime testing).

It's a fairly rushed commit for the POPL deadline and hopefully doesn't cause too much carnage.

### DIFF
--- a/runtime/libcn/dune
+++ b/runtime/libcn/dune
@@ -5,7 +5,7 @@
        (:src (glob_files src/*.c)))
  (action
   (progn
-   (run cc -Iinclude/ -c %{src})
+   (run cc -Iinclude/ -c -g %{src})
    (run ar -rcs %{target} alloc.o hash_table.o utils.o))))
 
 (install
--- a/runtime/libcn/src/utils.c
+++ b/runtime/libcn/src/utils.c
@@ -108,14 +108,18 @@ void ghost_stack_depth_decr(void) {
     // TODO - revisit this Dhruv over-confident change
     hash_table_iterator it = ht_iterator(cn_ownership_global_ghost_state);
     printf("CN pointers leaked at (%ld) stack-depth: ", cn_stack_depth);
+    _Bool leaked = false;
     while (ht_next(&it)) {
         uintptr_t *key = it.key;
         int *depth = it.value;
         _Bool fine = *depth <= cn_stack_depth;
-        if (!fine) printf(FMT_PTR_2 " (%d),", *key, *depth);
-        // c_ghost_assert(convert_to_cn_bool(fine));
+        if (!fine) {
+            leaked = true;
+            printf(FMT_PTR_2 " (%d),", *key, *depth);
+        }
     }
     printf("\n");
+    c_ghost_assert(convert_to_cn_bool(!leaked));
 }
 
 int ownership_ghost_state_get(signed long *address_key) {
@@ -175,7 +179,6 @@ void cn_put_ownership(uintptr_t generic_c_ptr, size_t size) {
     for (int i = 0; i < size; i++) { 
         signed long *address_key = alloc(sizeof(long));
         *address_key = generic_c_ptr + i;
-        /* printf(" off: %d [" FMT_PTR_2 "] (function: %s)\n", i, *address_key, error_msg_info.function_name); */
         int curr_depth = ownership_ghost_state_get(address_key);
         if (curr_depth != cn_stack_depth) {
             printf("CN memory access failed: function %s, file %s, line %d\n", error_msg_info.function_name, error_msg_info.file_name, error_msg_info.line_number);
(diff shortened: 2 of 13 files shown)


runtime/libcn/src/utils.c
```diff
@@ -96,10 +96,30 @@ void ghost_stack_depth_incr(void) {
...
 void ghost_stack_depth_decr(void) {
     cn_stack_depth--;
     // update_error_message_info(0);
     print_error_msg_info();
+    // TODO - revisit this Dhruv over-confident change
+    hash_table_iterator it = ht_iterator(cn_ownership_global_ghost_state);
+    printf("CN pointers leaked at (%ld) stack-depth: ", cn_stack_depth);
+    _Bool leaked = false;
+    while (ht_next(&it)) {
+        uintptr_t *key = it.key;
```
rbanerjee20: Can you change this to an `intptr_t`? The type of `it.key` is `signed long` and it's throwing a warning when building CN

runtime/libcn/src/utils.c
```diff
@@ -96,10 +96,30 @@ void ghost_stack_depth_incr(void) {
...
+        int *depth = it.value;
+        _Bool fine = *depth <= cn_stack_depth;
+        if (!fine) {
+            leaked = true;
+            printf(FMT_PTR_2 " (%d),", *key, *depth);
+        }
+    }
+    printf("\n");
+    c_ghost_assert(convert_to_cn_bool(!leaked));
```
rbanerjee20, line 122: The unfortunate issue with the current assertion setup is that there is a `cn_assert` function for asserting whether facts about the _CN runtime spec_ are true, and a distinct `c_ghost_assert` function for asserting whether facts about _C ownership_ are true or not. The printed error messages differ depending on which one is called. This falls apart when talking about the global ghost state as a whole, since the hash map `ownership_ghost_state` is currently being used for both types of ownership checking. However, I think it's alright to keep this in for now, and this should probably be a future discussion point regarding the design of the global ghost state and what it should store about the addresses it has mapped.

runtime/libcn/dune
```diff
@@ -5,7 +5,7 @@
        (:src (glob_files src/*.c)))
  (action
   (progn
-   (run cc -Iinclude/ -c %{src})
+   (run cc -Iinclude/ -c -g %{src})
```
rbanerjee20, line 8: Do we always want debug information for the runtime library itself? I feel like it's sufficient to just compile the CN-annotated files with `-g`

runtime/libcn/src/utils.c
```diff
@@ -428,6 +446,19 @@ void *cn_calloc(size_t num, size_t size)
...
+// TODO - revisit this Dhruv over-confident change *)
+void cn_free_sized(void* malloced_ptr, size_t size)
+{
+  printf("C ownership checking: remove malloc-ed:" FMT_PTR ", size: %lu\n", (uintptr_t) malloced_ptr, size);
+  for (int i = 0; i < size; i++) {
+      signed long *address_key = alloc(sizeof(long));
+      *address_key = (uintptr_t) malloced_ptr + i;
+      /* printf(" off: %d [" FMT_PTR "]\n", i, *address_key); */
+      int curr_depth = ownership_ghost_state_get(address_key);
+      c_ghost_assert(convert_to_cn_bool(curr_depth == cn_stack_depth));
```
rbanerjee20, line 457: should the free function itself be asserting anything about ownership? surely that is what the runtime CN spec is for, so this might be stronger than necessary

tests/cn/leaky_main.runtime.c
```diff
@@ -0,0 +1,29 @@
...
+  *p = x;
+  return p;
+}
+
+int main()
+/*@ trusted; @*/
+{
+    unsigned int *p = mkref(5);
+    /*@ assert (*p == 5u32); @*/
+    // TODO - consider decrementing stack-depth for main too?
```
rbanerjee20, line 25: Can consider, although right now we are treating `main` as slightly 'special' - the addresses of C globals currently are added to the ghost state here since the stack depth is assumed to be 0, although I suppose that's not a huge change if we decide to add the stack depth increment and decrement here too

runtime/libcn/src/utils.c
dc-mak: I'll add this as a test.
```c
int *malloc_int()
/*@ trusted; ensures take X = Block<int>(return); @*/
{
   return malloc(sizeof(int));
}

void bad_free(int *p)
/*@ trusted; ensures true; @*/
{
    cn_free_size(p, sizeof(int));
}

int main()
/*@ trusted; @*/
{
    int * p = malloc_int();
    bad(p):
}
```

runtime/libcn/dune
dc-mak: Yes - you often want to set the break point at `cn_exit_aux` and you can't do that with debug info.

rbanerjee20 (approved):

PR #379, opened 2024-07-10
Add numeric separator `_`

septract:
Addresses #337

PR #388, opened 2024-07-16
Allow ' in numeric constants

MedranoApol:
Allowing `'` in numeric constants, so you can write things like `34'i32`.

This adds this as an option, however `34i32` still works.


dc-mak:
Hi Apol, thanks, but this will be handled elsewhere see [this](https://github.com/rems-project/cerberus/pull/379) and [this](https://github.com/rems-project/cerberus/issues/337) this will be handled by a different PR. 

Also the feature is to allow them at any point in the digits like `300'00'000i32` not just before the type annotation.

PR #394, opened 2024-07-17
[CN] Use subcommands (`cn verify` and `cn generate-tests`)

ZippeyKeys12:
Meant to address #343.

Currently:
- Moves the current functionality under `cn verify` and deprecates the current use `cn [FILE]`.
- Splits off test generation under `cn generate-tests`.
- Changes a few flags to use `-` as a separator instead of `_`, for consistency.

~~Easy to add:~~
- ~~Move the executable spec into a subcommand (I just couldn't come up with a great name)~~

Other thoughts:
- Are there any flags used by the frontend that are assigned outside the `frontend` function?
- Ideally, we'd also move lemma export into a subcommand, but it seems very tied to CN checking currently.

### DIFF
--- a/backend/cn/main.ml
+++ b/backend/cn/main.ml
@@ -125,6 +125,96 @@ let check_input_file filename =
     if not (ext ".c" || ext ".h") then
       CF.Pp_errors.fatal ("file \""^filename^"\" has wrong file extension")
 
+let with_well_formedness_check
+      (* CLI arguments *)
+      ~filename
+      ~macros
+      ~incl_dirs
+      ~incl_files
+      ~csv_times
+      ~log_times
+      ~astprints
+      ~use_peval
+      ~no_inherit_loc
+      ~magic_comment_char_dollar
+      (* Callbacks *)
+      ~handle_error
+      ~(f :
+        prog5 : unit Mucore.mu_file ->
+        ail_prog : CF.GenTypes.genTypeCategory A.ail_program ->
+        statement_locs : Cerb_location.t CStatements.LocMap.t ->
+        paused : _ Typing.pause ->
+        unit Resultat.t)
+  =
+  check_input_file filename;
+  let (prog4, (markers_env, ail_prog), statement_locs) =
+    handle_frontend_error
+      (frontend ~macros ~incl_dirs ~incl_files astprints ~do_peval:use_peval ~filename ~magic_comment_char_dollar)
+  in
+  Cerb_debug.maybe_open_csv_timing_file ();
+  Pp.maybe_open_times_channel
+    (match (csv_times, log_times) with
+     | (Some times, _) -> Some (times, "csv")
+     | (_, Some times) -> Some (times, "log")
+     | _ -> None);
+  try
+    let result =
+      let open Resultat in
+      let@ prog5 = Core_to_mucore.normalise_file ~inherit_loc:(not(no_inherit_loc)) (markers_env, snd ail_prog) prog4 in
+      print_log_file ("mucore", MUCORE prog5);
+      let paused = Typing.run_to_pause Context.empty (Check.check_decls_lemmata_fun_specs prog5) in
+      Result.iter_error handle_error (Typing.pause_to_result paused);
+      f ~prog5 ~ail_prog ~statement_locs ~paused
+    in
+    Pp.maybe_close_times_channel ();
+    Result.fold ~ok:(fun () -> exit 0) ~error:handle_error result
+  with
+  | exc ->
+    Pp.maybe_close_times_channel ();
+    Cerb_debug.maybe_close_csv_timing_file_no_err ();
+    Printexc.raise_with_backtrace exc (Printexc.get_raw_backtrace ())
+
+let handle_type_error ~json ~state_file ~output_dir e =
+  if json then TypeErrors.report_json ?state_file ?output_dir e
+  else TypeErrors.report ?state_file ?output_dir e;
+  match e.msg with
+  | TypeErrors.Unsupported _ -> exit 2
+  | _ -> exit 1
+
+let well_formed
+      filename
+      macros
+      incl_dirs
+      incl_files
+      json
+      state_file
+      output_dir 
+      csv_times
+      log_times
+      astprints
+      use_peval
+      no_inherit_loc
+      magic_comment_char_dollar
+  =
+  with_well_formedness_check
+    ~filename
+    ~macros
+    ~incl_dirs
+    ~incl_files
+    ~csv_times
+    ~log_times
+    ~astprints
+    ~use_peval
+    ~no_inherit_loc
+    ~magic_comment_char_dollar
+    ~handle_error:(handle_type_error ~json ~state_file ~output_dir)
+    ~f:(fun
+        ~prog5:_
+        ~ail_prog:_
+        ~statement_locs:_
+        ~paused:_
+        -> Resultat.return ())
+
 let verify
       filename
       macros
@@ -139,7 +229,7 @@ let verify
       no_timestamps
       json
       state_file
-      output_dir 
+      output_dir
       diag
       lemmata
       only
@@ -168,7 +258,7 @@ let verify
         CF.Pp_errors.fatal ("debug level must be 0 for json output");
       if print_level > 0 then
         CF.Pp_errors.fatal ("print level must be 0 for json output");
-    end;
+  end;
   begin (*flags *)
     Cerb_debug.debug_level := debug_level;
     Pp.loc_pp := loc_pp;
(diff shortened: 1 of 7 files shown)


dc-mak:
Thanks for this, but perhaps I should have been clearer - I only wanted `cn generate-tests` as a new subcommand, but not `cn verify`. This is to make it easier to transition the CI in the multiple places it's used currently. Is that possible, or is it only possible to have `generate-tests` if there's another subcommand for the rest?

backend/cn/main.ml
```diff
@@ -499,12 +501,58 @@ let () =
...
+  verify_cmd;
+  generate_tests_cmd;
+]
+
+let () =
   let version_str = "CN version: " ^ Cn_version.git_version in
   let cn_info = Cmd.info "cn" ~version:version_str in
-  Stdlib.exit @@ Cmd.(eval (v cn_info check_t))
-
+  Stdlib.exit @@ Cmd.(eval (group cn_info subcommands))
```
dc-mak: I think if you put `verify_cmd` as the default instead of `cn_info` this should retain backwards compatibility - with a deprecation warning and suggestion to use `verify`.
You can keep `verify_cmd` in the subcommands list so that both work, and we can transition the CI smoothly.

backend/cn/main.ml
```diff
@@ -246,6 +234,35 @@ let main
...
+    handle_frontend_error
+      (frontend ~macros ~incl_dirs ~incl_files astprints ~do_peval:use_peval ~filename ~magic_comment_char_dollar)
+  in
+  let (_, sigma) = ail_prog in
+  let _ =
+    let open Resultat in
+    let@ prog5 = Core_to_mucore.normalise_file ~inherit_loc:(not(no_inherit_loc)) (markers_env, snd ail_prog) prog4 in
+    Cerb_colour.without_colour (fun () ->
+      TestGeneration.main ~output_dir ~filename ~max_unfolds sigma prog5 testing_framework)
+    ();
```
dc-mak: Please run the specification well-formedness checking before you do test generation (see the main for the executable spec)
`let paused = Typing.run_to_pause Context.empty (Check.check_decls_lemmata_fun_specs prog5) in ...`

backend/cn/main.ml
```diff
@@ -421,49 +438,34 @@ module Executable_spec_flags = struct
...
 
 let with_ownership_checking =
   let doc = "Enable ownership checking within CN runtime testing" in
-  Arg.(value & flag & info ["with_ownership_checking"] ~doc)
+  Arg.(value & flag & info ["with-ownership-checking"] ~doc)
 
 let copy_source_dir =
   let doc = "Copy non-CN annotated files into output_decorated_dir for CN runtime testing" in
-  Arg.(value & flag & info ["copy_source_dir"] ~doc)
+  Arg.(value & flag & info ["copy-source-dir"] ~doc)
```
dc-mak, line 543: @rbanerjee20 Can you check that all the scripts in the cerberus repo which use these flags have been covered?

backend/cn/main.ml
```diff
@@ -361,7 +378,7 @@ let csv_times =
 
 let log_times =
   let doc = "file in which to output hierarchical timing information" in
-  Arg.(value & opt (some string) None & info ["log_times"] ~docv:"FILE" ~doc)
+  Arg.(value & opt (some string) None & info ["log-times"] ~docv:"FILE" ~doc)
```
dc-mak: @cp526 is this used in any scripts anywhere?

dc-mak (changes requested):
Thanks so much for this! Just few changes so that CI in the CN tutorial can be transitioned smoothly too.

backend/cn/main.ml
ZippeyKeys12: That makes sense

backend/cn/main.ml
cp526: Don't think so.

backend/cn/main.ml
rbanerjee20: sure, can update them after Zain has merged since they are not part of the Cerberus CI yet?

dc-mak:
To reply:

> Move the executable spec into a subcommand (I just couldn't come up with a great name)

@rbanerjee20 will move the executable spec stuff into a subcommand in a separate commit (and come up with a good verb for it)

> Are there any flags used by the frontend that are assigned outside the frontend function?

I don't believe so (we can always add later)

> Ideally, we'd also move lemma export into a subcommand, but it seems very tied to CN checking currently.

Yeah leave it for now, we can revisit later. Goal is to make things better and easier for the future rather than to make it perfect.

backend/cn/main.ml
ZippeyKeys12: How should I implement `handle_error`? I'd be duplicating quite a lot...

dc-mak: Hmmm. Lift it out and move any flags it needs to `Common_flags`. You may wish to use the common `output-dir` flag for your test generation too. In fact, you may need to factor out everything up to that point, rather than just the handle error.

backend/cn/main.ml
```diff
@@ -421,90 +438,138 @@ module Executable_spec_flags = struct
...
   in
-  let version_str = "CN version: " ^ Cn_version.git_version in
-  let cn_info = Cmd.info "cn" ~version:version_str in
-  Stdlib.exit @@ Cmd.(eval (v cn_info check_t))
+  let doc = String.trim
+    "
+    Generates RapidCheck tests for all functions in [FILE] with CN specifications.
+    The tests use randomized inputs, which are guaranteed to satisfy the CN precondition.
+    A [.cpp] file containing the test harnesses will be placed in [output-dir].
+    "
```
dc-mak: This can use quoted string literals or escaped newlines instead of trimming https://ocaml.org/manual/5.2/lex.html#sss:stringliterals

backend/cn/main.ml
```diff
@@ -421,90 +438,138 @@ module Executable_spec_flags = struct
...
 
+let () =
+  (** Minor hack, as Cmdliner [default] doesn't work with positional arguments *)
+  let using_subcommands =
+    if Array.length Sys.argv >= 2
+    then
+      let first_arg = Sys.argv.(1) in
+      subcommands
+      |> List.find_opt (fun cmd -> String.equal (Cmd.name cmd) first_arg)
+      |> Option.is_some
```
dc-mak: `List.exists`

dc-mak (changes requested):

backend/cn/main.ml
ZippeyKeys12: Okay, I just print the message and point the user to `cn verify`

tests/cn-runtime/cn-runtime-dir.sh
```diff
@@ -37,7 +37,7 @@ do
   fi
   EXEC_C_FILE=$EXEC_C_DIRECTORY/$TEST_BASENAME-exec.c
   echo Generating $EXEC_C_FILE ...
-  if ! cn $TEST_NAME --output_decorated=$TEST_BASENAME-exec.c --output_decorated_dir=$EXEC_C_DIRECTORY/ --with_ownership_checking
+  if ! cn verify $TEST_NAME --output-decorated=$TEST_BASENAME-exec.c --output-decorated-dir=$EXEC_C_DIRECTORY/ --with-ownership-checking
```
rbanerjee20, line 40: is this just until there's a subcommand for executable spec and afterwards `verify` will be replaced with whatever subcommand that is?

rbanerjee20, line 40: is this just until there is a subcommand for runtime testing?

dc-mak: To make the CI transition easier, `cn` and `cn verify` should work the same but just `cn` should print a deprecation warning of some sort. So... yes? Though it can be changed to `cn $TEST_NAME` .

dc-mak: To make the CI transition easier, `cn` and `cn verify` should work the same but just `cn` should print a deprecation warning of some sort. So... yes? Though it can be changed to `cn $TEST_NAME` .

backend/cn/main.ml
```diff
@@ -246,6 +234,54 @@ let main
...
+  let (prog4, (markers_env, ail_prog), _) =
+    handle_frontend_error
+      (frontend ~macros ~incl_dirs ~incl_files astprints ~do_peval:use_peval ~filename ~magic_comment_char_dollar)
+  in
+  let (_, sigma) = ail_prog in
+  try
+    let handle_error (e : TypeErrors.type_error) =
+      let report = TypeErrors.pp_message e.msg in
+      Pp.error e.loc report.short (Option.to_list report.descr);
+      print_string ("\nRun `cn verify " ^ filename ^ "` for more details\n");
```
dc-mak: Please delete; `cn verify` won't give more details.

backend/cn/main.ml
```diff
@@ -246,6 +234,54 @@ let main
...
+      ();
+      return ()
+    in
+    Pp.maybe_close_times_channel ();
+    Result.fold ~ok:(fun () -> exit 0) ~error:handle_error result
+  with
+  | exc ->
+    Pp.maybe_close_times_channel ();
+    Cerb_debug.maybe_close_csv_timing_file_no_err ();
+    Printexc.raise_with_backtrace exc (Printexc.get_raw_backtrace ())
```
dc-mak: This is a lot of code duplication. Is it possible to factor out the bits common to the existing `main` and generate tests?

dc-mak (changes requested):

backend/cn/main.ml
ZippeyKeys12: Uhh, so I did this, but it's not the most beautiful thing, so I can revert it.

A side effect is that it was trivial to add a `cn check` which just does the well-formedness check (I was thinking along the lines of `cargo check`, though I can remove it)

dc-mak: I think `cn lint-spec` might be more accurate since `check` and `verify` sound a little too similar.

backend/cn/main.ml
```diff
@@ -421,90 +527,164 @@ module Executable_spec_flags = struct
...
+    Verify_flags.output_dir $
+    Common_flags.csv_times $
+    Common_flags.log_times $
+    Common_flags.astprints $
+    Common_flags.use_peval $
+    Common_flags.no_inherit_loc $
+    Common_flags.magic_comment_char_dollar
+  in
+  let doc = "Run CN well-formedness check" in
+  let info = Cmd.info "check" ~doc in
```
dc-mak: Can you call it `lint-spec`? Just because the distinction between `verify` and `check` may be confusing (we can always change it later, but for now).

dc-mak (changes requested):
Alright, one final thing and it's good to go - please make the change, squash it all into one commit, and mark the PR as not a draft. Thank you very much for your patience and putting this together - much appreciated! 👍

backend/cn/main.ml
cp526: But `lint-spec` is not what it does. Perhaps `cn wellformed`, even if that's not a verb.

ZippeyKeys12: What about `cn sanity-check`?

backend/cn/main.ml
ZippeyKeys12: What about `cn sanity-check`? `cn integrity`? `cn form-check`? `cn wf` but with a detailed `doc`?

tests/cn-runtime/cn-runtime-dir.sh
rbanerjee20: Yes, that’s all I meant. This is the runtime testing script so having the `verify` command in there seems wrong. I’ll add the subcommand soon enough that this deprecation problem shouldn’t happen for very long.

backend/cn/main.ml
rbanerjee20: I like the last one

tests/cn-runtime/cn-runtime-dir.sh
ZippeyKeys12: You should just have to figure out which flags you care about and call `with_well_formedness_check`.

backend/cn/main.ml
ZippeyKeys12: It's also my favorite, so I changed `check` to `wf`. It being opaque stops users from assuming its functionality.

PR #397, opened 2024-07-18
Expand `let` in results of the solver.

### LINKED ISSUE #395: [CN] Unexpected solver error
```c
typedef unsigned long int uintptr_t;

unsigned char* owned_int_ptr_to_owned_uchar_arr(int **addr_p)
/*@
trusted;

requires
    take P = Owned<int*>(addr_p);

ensures
    return == addr_p;
    take B = Owned<unsigned char[(sizeof(int*))]>(return);
    (u64) P == shift_left((u64)B[7u64], 56u64)
                + shift_left((u64)B[6u64], 48u64)
                + shift_left((u64)B[5u64], 40u64)
                + shift_left((u64)B[4u64], 32u64)
                + shift_left((u64)B[3u64], 24u64)
                + shift_left((u64)B[2u64], 16u64)
                + shift_left((u64)B[1u64], 8u64)
                + (u64)B[0u64];
@*/
{
    return (unsigned char*)addr_p;
}


int x=1;
int main()
/*@ accesses x; @*/
{
  int *p = &x+1;
  unsigned char *p_bytes = owned_int_ptr_to_owned_uchar_arr(&p);
}
```
and 
```zsh
15:45 ➜  cerberus git:(master) cn unexpected_sovler.error.c
unexpected_sovler.error.c:11:5: warning: CN pointer equality is not the same as C's (will not warn again). Please use `ptr_eq` or `is_null` (maybe `addr_eq`).
    return == addr_p;
    ~~~~~~~^~~~~~~~~
[1/1]: main
cn: internal error, uncaught exception:
    UnexpectedSolverResponse(((a!1
      (store
       (store
        (store ((as const (Array (_ BitVec 64) (_ BitVec 8))) #x60)
         #x0000000000000005 #x45)
        #x0000000000000000 #x14)
       #x0000000000000007 #x46))))
    Raised at Dune__exe__Simple_smt.to_con.bad in file "backend/cn/simple_smt.ml", line 728, characters 15-51
    Called from Dune__exe__Solver.get_value in file "backend/cn/solver.ml", line 370, characters 18-33
    Called from Dune__exe__Solver.get_ivalue in file "backend/cn/solver.ml" (inlined), line 355, characters 8-33
    ...
```

yav:
Fixes #395.  It would appear that `z3` will return terms containing `let` as a result of `get-value`.


cp526:
Thanks for the fixes. One comment inline below.

backend/cn/simple_smt.ml
```diff
@@ -662,6 +662,44 @@ let get_expr s v: sexp =
...
+    in
+    let su = List.fold_left add_bind StrMap.empty binds in
+    Some (su, body)
+  | _ -> None
+
+(** Expand let-definitions in this term.
+ NOTE: this is intended to be used mostly on models generated by the
+ solver (e.g., `get-value` in Z3 sometimes contains `let`).  As such
+ we assume that `forall` and `exist` won't occur, and so we don't need
+ to check for variable capture. *)
```
cp526, line 682: Won't the smt solver potentially return lambdas as models for arrays or functions, which we'd have to be careful about regarding variable capture?

cp526: @yav

yav: It's hard to say as there is no standard for this, but in the cases I've seen arrays seem to be rendered as a bunch of concrete updates to a constant array.   It's probably worth trying to see if we can make it break though.  I am not even sure what notation the solvers might use for lambdas, the "standard" binders I see are `let`, `forall`, `exists`, and, I guess, the patterns in `match`, so I was thinking that I'd only expect to see `let` in a model.

cp526: I've experimented locally a bit, but so far have failed to get z3 or cvc5 to produce a lambda in a countermodel. (Though I have seen some `define-fun`s for function, and perhaps array, countermodels.) 

If you think we're good to go regarding binders (?) I'm happy to merge.

yav: I think we should merge it as is.   Admittedly, it would be a nasty bug to track down if this happens, but it is hard to fix it (or guard against it) without knowing what might be the binding construct...   I am pretty sure that `forall` and `exist` don't make sense in this context.

I see we have merge conflicts, so I'll resolve them, and then we can merge

yav:
I am going to close this and make a new version, staring post reformatting and moving things around.

PR #400, opened 2024-07-19
CN: Another round of CN tidying (new PR)

bcpierce00:


backend/cn/solver.mli
```diff
@@ -1,25 +1,36 @@
+(* Module Solver -- Interface to the SMT solver via SMTLIB *)
 type solver
 type model
+(* (TODO: BCP: The "with quantifiers" part will be the instantiations that the
+   solver found -- is that right?) *)
```
cp526, line 5: Yes. To implement `provable (Forall ((s,bt), t))`, CN picks a fresh variable `s'` and checks that `not t[s'/s]` is unsatisfiable ("there exists no choice of s such that t doesn't hold"). If `provable` returns false, i.e. the SMT solver found an `s'` for which `t[s'/s]` doesn't hold, then `model_with_q` has that `s'` for its second component.

backend/cn/solver.mli
```diff
@@ -1,25 +1,36 @@
...
-val pop : solver -> int -> unit
+(* Incrementally (and imperatively) add an assumption to the solver state *)
 val add_assumption : solver -> Global.t -> LogicalConstraints.t -> unit
 
-val set_slow_smt_settings : float option -> string option -> unit
+(* Save / restore solver state, to support backtracking *)
+val push : solver -> unit
+val pop : solver -> int -> unit
 
+(* TODO: BCP: What is this? *)
```
cp526, line 28: `num_scopes` is the count of the open `push` scopes.

backend/cn/solver.mli
```diff
@@ -1,25 +1,36 @@
...
+val num_scopes : solver -> int
 
+(* Run the solver.  Note that we pass the assumptions explicitly even
+   though they are also available in the solver context, because CN is
+   going some simplification on its own. *)
```
cp526, line 33: Plus we have a default quantifier instantiation heuristic, which adds additional constraints derived from `assumptions` based on the to-be-proved constraint.

backend/cn/solver.mli
```diff
@@ -29,24 +40,28 @@ val provable :
...
+(* Ask the solver for the model that it found in a call to [provable] *)
 val model :
   unit ->
   model_with_q
 
-
-
+(* Ask the solver to evaluate a CN term in the context of a
+   model.  (Might return None in case we ask for the value of a "don't
+   care" value in the (minimal) model.) *)
```
cp526, line 50: This is in preparation for that, but minimal models aren't implemented yet.

backend/cn/solver.mli
```diff
@@ -29,24 +40,28 @@ val provable :
...
+(* TODO: BCP: I don't understand how this could ever be called -- how do we get a
+   model to pass it??? *)
```
cp526, line 52: We would have called the `model` function from above

backend/cn/solver.mli
```diff
@@ -29,24 +40,28 @@ val provable :
...
 val eval :
-  Global.t ->
+  Global.t ->     (* TODO: BCP: IIUC Christopher thinks this is not needed? *)
```
cp526, line 54: Here and in other places in the solver module, I think, the `global` argument can be removed: the solver is initialised with a `global` typing context at some point earlier on, and then this is expected not to change any longer.

backend/cn/solver.mli
```diff
@@ -29,24 +40,28 @@ val provable :
...
   model ->
   IndexTerms.t ->
   IndexTerms.t option
 
+(* TODO: BCP: What is this? *)
+val set_slow_smt_settings : float option -> string option -> unit
```
cp526, line 60: Not sure if that's currently being used. We used to have an option to specifically log slow-running SMT queries.

backend/cn/solver.mli
```diff
@@ -29,24 +40,28 @@ val provable :
...
 
-
-
+(* Debugging *)
+(* TODO: BCP: This one seems misnamed -- it doesn't return a string...? *)
```
cp526, line 63: Agreed.

backend/cn/symRel.mli
```diff
@@ -1,89 +0,0 @@
-(* Module SymPair -- TODO: BCP: What is it for?? *)
-
-(* TODO: BCP: Some filenames use_underscores, while some areInCamlCase
-   -- any compelling reason for that? *)
```
cp526, line 4: None at all. Originally everything (in CN at least) was CamlCase, then other people added different names.

cp526:
Thanks very much!

cp526:
I added some comments for some of the todos. Some module interfaces should be shrunk, but we can do that after ocamlformat is in place.

bcpierce00:
Thanks.

For future PRs like this, please feel free to add such comments inline in
the .ml files -- that makes sure they don't get lost and also makes it easy
for you to respond to some queries by directly adding or fixing what needs
to be added or fixed, saving the step of transcription from github to ml
files later.

PR #403, opened 2024-07-22
CN: Format code

dc-mak:
These commits auto format the backend/cn directory and add formatting checks to the CI. It excludes pre-commit hooks since those are done locally and so need to done on a per-user basis.

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,2 @@
+# CN: Format code according to .ocamlformat file
+36882ab6e66ebb5e8c36e2bb7b80dbfa5be64322
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -91,7 +91,7 @@ jobs:
         opam switch ${{ matrix.version }}
         eval $(opam env --switch=${{ matrix.version }})
         opam pin --yes --no-action add cn .
-        opam install --yes cn
+        opam install --yes cn ocamlformat.0.26.2
 
     - name: Checkout cn-tutorial
       uses: actions/checkout@v4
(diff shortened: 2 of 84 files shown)


bcpierce00:
A few little things I noticed...

* I wish it would conserve space a bit more when things like pattern branches will fit all on one line.
image

* There are many places where the formatter leaves two blank lines intead of one between definitions, for reasons I do not understand (I don't see the pattern).

* The thing it did on line 22 here
image
seems bizzarre and wrong -- replacing a specific pattern with a simple variable.  (In this particular case there's no particular harm, but with a datatype other than list it could create a bug later when someone adds a new constructor and the compiler doesn't signal that this match needs to be updated.)

cp526:
> A few little things I noticed...
> 
> * I wish it would conserve space a bit more when things like pattern branches will fit all on one line.
>   image

Agreed. Though, when setting up the `.ocamlformat` file I've been looking at the options for pattern-match formatting, and IIRC I didn't find an option that does the compact formatting in your example at simultaneously *doesn't put multiple pattern match branches on the same line* when those fit (which typically is bad for readability).

> * There are many places where the formatter leaves two blank lines intead of one between definitions, for reasons I do not understand (I don't see the pattern).
> * The thing it did on line 22 here
>   image
>   seems bizzarre and wrong -- replacing a specific pattern with a simple variable.  (In this particular case there's no particular harm, but with a datatype other than list it could create a bug later when someone adds a new constructor and the compiler doesn't signal that this match needs to be updated.)

Luckily those seem to be two different functions: `mk_arg0` vs `mk_arg1`.

bcpierce00:
> Agreed. Though, when setting up the .ocamlformat file I've been looking at the options for pattern-match formatting, and IIRC I didn't find an option that does the compact formatting in your example at simultaneously doesn't put multiple pattern match branches on the same line when those fit (which typically is bad for readability).

If I have to choose, I think I'd rather live with multiple short branches on the same line.

> Luckily those seem to be two different functions: mk_arg0 vs mk_arg1. 

Oh yes, I was reading too fast!

dc-mak:
Pick your poison.
```diff
@@ -976,11 +897,7 @@ end = struct
         let@ ftyp =
           RI.General.ftyp_args_request_step rt_subst loc uiinfo original_resources ftyp
         in
-        match ftyp with
-        | I rt ->
-          return rt
-        | _ ->
-          check ftyp
+        match ftyp with I rt -> return rt | _ -> check ftyp
       in
```

bcpierce00:
If these are the only choices I strongly prefer the second.

PR #405, opened 2024-07-22
[CN] Add modulo operator `%`

MedranoApol:


elaustell:
I think this looks great Apol! I have no suggestions except to squash.

dc-mak (changes requested):
Lgtm, please rebase and squash if necessary and I'll merge asap.

dc-mak (changes requested):
Code needs formatting too - this was missed the first time because the pipelines stopped at the failing test (a file which tested for a crash on modulus, so that's expected). The cn-tutorial has been updated to remove the failing test temporarily and so after formatting, this should all be green.

PR #415, opened 2024-07-23
CN: Remove deprecated command format

dc-mak:
There's a slight bug in Cmdliner, if the version string ends with a ')' then it's deleted (in `cn --help`), so this commit also changes the version string to end with a ']' instead.


dc-mak:
Last step for this too: https://github.com/rems-project/cerberus/issues/343

PR #421, opened 2024-07-23
[CN-Exec] Use `output-decorated-dir` as directory by using `Filename.concat`

ZippeyKeys12:
I assume this is the intended behavior ("to the provided directory" from `--help`).

If not, I think the `doc` for the `output-decorated-dir` should be a bit clearer.


PeterSewell (commented):
This is a bit hacky - really the subsequent code should use OCaml standard library Filename functions, eg `Filename.concat` instead of plain string concatenation `^`.

PR #429, opened 2024-07-24
CN: split CN into a library and an executable

### LINKED ISSUE #315: [CN] Repackage CN as separate library and executable
Currently, CN is packaged as an executable (in `dune` terms). As far as I can tell, this makes it impossible for other OCaml packages/projects to import/rely on CN modules, because (again, as far as I can tell - my OCaml/`opam`/`dune` is rusty) projects can only depend on libraries, not executables. Assuming the choice of this shape of packaging was arbitrary, I think we ought to reorganize things such that other projects can depend on CN, and can import the bulk of CN's modules.

One approach could look something like:
- Create two new directories, `backend/cn/{lib,bin}`
- Move `backend/cn/main.ml` to `backend/cn/bin`
  - Create an appropriate `backend/cn/bin/dune` file there
- Move everything else in `backend/cn` (except, perhaps, `manual`) to `backend/cn/lib`
  - Revise the just-moved `dune` file to declare itself as a `library` rather than an `executable`
- Make a trivial `backend/cn/dune` file that points to the two subdirectories

Some local experimentation suggests this approach is viable. Though a directory shuffle ought to have negligible impact on CN development in general, I realize it would be a bit painful immediately, especially for anyone with unmerged CN changes lying around. How does this land with folks? Does it seem like a good idea, and/or does it seem burdensome?

samcowger:
See #315 for context, but the short version is that CN is currently packaged as an executable, which means that other OCaml projects can't depend on it as a library. This PR packages almost all of CN's existing functionality as a library.

Fixes #315.


samcowger:
Interested parties probably include, but are not limited to, @cp526, @dc-mak, and @kmemarian.

cp526:
That looks good to me. 

How do we control what is exposed as library functionality, is that by writing `mli` files?

cp526:
This would be good to merge as soon as possible. 

Other than the inadvertent conflicts that have to be resolved, we only have to decide on how to sequence this wrt to any other inflight changes.

backend/cn/lib/core_to_mucore.ml
```diff
@@ -32,7 +32,7 @@ module SymSet = Set.Make (Sym)
 module SymMap = Map.Make (Sym)
 module IdMap = Map.Make (Id)
 open Print
-open Cn
+open CF.Cn
```
dc-mak: It's possible to get rid of the open altogether (though that may be orthogonal to this PR). The changes of to `CF.Cn` can then also be reverted to minimise the diff.
```diff
13:06 ➜  cerberus git:(c76fe3d9f) ✗ git --no-pager diff --
diff --git a/backend/cn/lib/core_to_mucore.ml b/backend/cn/lib/core_to_mucore.ml
index 7f406095c..67b522fca 100644
--- a/backend/cn/lib/core_to_mucore.ml
+++ b/backend/cn/lib/core_to_mucore.ml
@@ -4,6 +4,7 @@ open Lem_pervasives
 open Ctype
 open Milicore
 module CF = Cerb_frontend
+module Cn = CF.Cn
 module CA = Cabs_to_ail
 module CAE = Cabs_to_ail_effect
 open Cerb_frontend.Pp_ast
@@ -32,7 +33,6 @@ module SymSet = Set.Make (Sym)
 module SymMap = Map.Make (Sym)
 module IdMap = Map.Make (Id)
 open Print
-open CF.Cn
 open Resultat

 open Effectful.Make (Resultat)
@@ -1238,7 +1238,7 @@ let add_spec_arg_renames
   loc
   args
   arg_cts
-  (spec : (Symbol.sym, Ctype.ctype) cn_fun_spec)
+  (spec : (Symbol.sym, Ctype.ctype) Cn.cn_fun_spec)
   env
   =
   List.fold_right
@@ -1248,7 +1248,7 @@ let add_spec_arg_renames
         fun_sym
         (Memory.sbt_of_sct (convert_ct loc ct))
         env)
-    (List.combine args (List.combine arg_cts spec.cn_spec_args))
+    (List.combine args (List.combine arg_cts spec.Cn.cn_spec_args))
     env


@@ -1318,13 +1318,13 @@ let normalise_fun_map_decl
                         ^^ P.colon
                         ^^^ !^"re-specification of CN annotations from:"
                         ^^ P.break 1
-                        ^^^ Locations.pp spec.cn_spec_loc)
+                        ^^^ Locations.pp spec.Cn.cn_spec_loc)
                  }
            in
            let env = add_spec_arg_renames loc args (List.map snd arg_cts) spec env in
            let env =
              C.add_renamed_computational
-               spec.cn_spec_ret_name
+               spec.Cn.cn_spec_ret_name
                ret_s
                (Memory.sbt_of_sct (convert_ct loc ret_ct))
                env
@@ -1382,7 +1382,7 @@ let normalise_fun_map_decl
        return (Some (M_Proc (loc, args_and_body, trusted, desugared_spec), mk_functions))
      | Mi_ProcDecl (loc, ret_bt, _bts) ->
        (match SymMap.find_opt fname fun_specs with
-        | Some (_ail_marker, (spec : (Symbol.sym, Ctype.ctype) cn_fun_spec)) ->
+        | Some (_ail_marker, (spec : (Symbol.sym, Ctype.ctype) Cn.cn_fun_spec)) ->
           let@ () =
             check_against_core_bt loc ret_bt (Memory.bt_of_sct (convert_ct loc ret_ct))
           in
@@ -1554,7 +1554,7 @@ let register_glob env (sym, glob) =

 (* |> C.add_c_var_value sym (IT.sym_ (sym, bt)) *)

-let translate_datatype env { cn_dt_loc; cn_dt_name; cn_dt_cases; cn_dt_magic_loc = _ } =
+let translate_datatype env Cn.{ cn_dt_loc; cn_dt_name; cn_dt_cases; cn_dt_magic_loc = _ } =
   let translate_arg (id, bt) =
     (id, SBT.to_basetype (Compile.translate_cn_base_type env bt))
   in
@@ -1585,7 +1585,7 @@ let normalise_file ~inherit_loc ((fin_markers_env : CAE.fin_markers_env), ail_pr
   let env = List.fold_left register_glob env globs in
   let fun_specs_map =
     List.fold_right
-      (fun (id, spec) acc -> SymMap.add spec.cn_spec_name (id, spec) acc)
+      (fun (id, spec) acc -> SymMap.add spec.Cn.cn_spec_name (id, spec) acc)
       ail_prog.cn_fun_specs
       SymMap.empty
   in

```

dc-mak (commented):
Lgtm, left an inline suggestion on `core_to_mucore.ml`.

dc-mak:
> This would be good to merge as soon as possible.
> 
> Other than the inadvertent conflicts that have to be resolved, we only have to decide on how to sequence this wrt to any other inflight changes.

Agreed, though almost all is file-level renamings, so shouldn't be too much of an issue for rebasing either which way: the `Cn` to `CF.Cn` might cause some merge conflicts, but none of them are currently under active dev and it would be minor/easy to resolve.

backend/cn/lib/core_to_mucore.ml
samcowger: In general, I certainly prefer fewer `open`s and more qualified names. My only concern would be someone seeing `Cn.x` here and thinking it was something local, rather than something from `Cerb_frontend`, but it isn't a strong concern. Happy to make the change if you like.

samcowger: Or perhaps I ought to defer to #433?

dc-mak: Yeah, I did the diff and then got carried away... It's all sorted now. The librarification won't cause any issues with it since the name `Cn` will be unambiguous to the build system.

cp526:
Thanks very much!

Issue #14, opened 2018-06-06
Stack overflow


victorgomes:
This is not necessarily an error, but it shows how easy it is to get an stack overflow in Cerberus.
If one sets the `j` to 1 or 2, it runs OK, otherwise stack overflow!

```c 
#include <stdlib.h>
#include <string.h>

typedef struct {
  int a[10];
} S;

int main()
{
  S* ptr_s1=malloc(15*sizeof(S));
  int j = 3;
  if(ptr_s1 != NULL) {
    for(int i=0;i<15;i++) {
      memset(ptr_s1,1,j*sizeof(S));
    }
  }
}
```

PR #430, opened 2024-07-24
Build multiplatform docker images

### LINKED ISSUE #431: Build multiplatform CN docker image
Currently only `linux/amd64` platform is supported, which works well for intel based hosts, but leads to slow performance on M1/M2 Macs (they are arm based).

Githu action runners are intel based, thus the ARM image needs to be done in emulation.

podhrmic:
@cp526 this should take care of multiplatform image builds. The non-native build takes a looong time (over 3 hours), so I split it into a separate file that runs only on merge to master.

Fixes https://github.com/rems-project/cerberus/issues/431

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -147,6 +147,9 @@ jobs:
 
   deploy-docker:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        platform: [linux/amd64, linux/arm64]
     permissions:
       packages: write
       contents: read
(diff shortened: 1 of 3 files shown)


podhrmic:
The build seems to be stuck at `#14 4097.7 -> installed ppx_sexp_conv.v0.17.0`

cp526:
Thanks! I'll check whether there's an ARM github runner available somewhere, but I suspect there won't be. Does Galois have one? There might be a way to set up Cerberus's CI to "ping" another repository which then builds the ARM image (on a faster machine)?

Absent that, I wonder whether, at 3 hours, the ARM docker image job takes too long to be practical. Would we be notified of CI test failures before the ARM job is finished (to ensure feedback is quick enough)?

podhrmic:
@cp526 on a second though, the 3 hour job is probably fine - I split when the jobs are run, and the docker images are now built only after a successful merge to `master`. The `build` CI job still runs after each commit. The docker job should never fail, as it basically runs the same steps as the `build` job.

A good next step would be to turn [this README](https://github.com/rems-project/cerberus/blob/master/backend/cn/README.md) into a script that can be called both from the `build` job and from the `Dockerfile`, ensuring an identical build config. Plus then the user can run the *same* script to build cerberus/cn. I am happy to do that, but I should probably wait until you have the cerberus/cn split done?

PR #432, opened 2024-07-25
CN: Tidy up in core_to_mucore.ml and signature

dc-mak:
In commenting on #429, I got sucked into tidying up `core_to_mucore.ml` and its signature.


bcpierce00:
Nice.

PR #447, opened 2024-07-30
[CN] Pure term (partial) evaluation

ZippeyKeys12:
Evaluates terms (`IndexTerms.t`). Can be used for term simplification.
It is needed to evaluate terms when generating test cases in OCaml-land.

Potential issues
- It's supposed to get stuck on errors, but it still might throw exceptions due to the functions used (ex: `List.combine`)
  - Some of these checks could be omitted, assuming the term is well-typed.

Other
- No sets, pointer ops, casts and a few other operations

### DIFF
--- a/backend/cn/lib/eval.ml
+++ b/backend/cn/lib/eval.ml
@@ -1,0 +1,607 @@
+module BT = BaseTypes
+module IT = IndexTerms
+module SymSet = Set.Make (Sym)
+
+let check_bits_bt (sgn1, sz1) (sgn2, sz2) here : (unit, string) result =
+  if not @@ BT.equal_sign sgn1 sgn2 then
+    Error ("Mismatched signs at " ^ Locations.simple_location here)
+  else if sz1 <> sz2 then
+    Error ("Mismatched sizes at " ^ Locations.simple_location here)
+  else
+    Ok ()
+
+
+let eval_num_binop
+  (eval : IT.t -> (IT.t, string) result)
+  (bt : BT.t)
+  (here : Locations.t)
+  (f : Z.t -> Z.t -> Z.t)
+  (it1 : IT.t)
+  (it2 : IT.t)
+  (loc : string)
+  : (IT.t, string) result
+  =
+  let ( let@ ) = Result.bind in
+  let return = Result.ok in
+  let@ it1 = eval it1 in
+  let@ it2 = eval it2 in
+  match (it1, it2) with
+  | IT (Const (Z n1), _, _), IT (Const (Z n2), _, _) ->
+    return @@ IT.num_lit_ (f n1 n2) bt here
+  | IT (Const (Bits ((sgn1, sz1), n1)), _, _), IT (Const (Bits ((sgn2, sz2), n2)), _, _)
+    ->
+    let@ () = check_bits_bt (sgn1, sz1) (sgn2, sz2) here in
+    return @@ IT.num_lit_ (BT.normalise_to_range_bt bt (f n1 n2)) bt here
+  | _, IT (Const (Z _), _, _) ->
+    Error ("Not constant integer `" ^ Pp.plain (IT.pp it1) ^ "` (" ^ loc ^ ")")
+  | _, IT (Const (Bits _), _, _) ->
+    Error ("Not constant bitvector `" ^ Pp.plain (IT.pp it1) ^ "` (" ^ loc ^ ")")
+  | IT (Const (Z _), _, _), _ ->
+    Error ("Not constant integer `" ^ Pp.plain (IT.pp it2) ^ "` (" ^ loc ^ ")")
+  | IT (Const (Bits _), _, _), _ ->
+    Error ("Not constant bitvector `" ^ Pp.plain (IT.pp it2) ^ "` (" ^ loc ^ ")")
+  | _, _ ->
+    Error
+      ("Not constant integer or bitvector `"
+       ^ Pp.plain (IT.pp it1)
+       ^ "` and `"
+       ^ Pp.plain (IT.pp it2)
+       ^ "` ("
+       ^ loc
+       ^ ")")
+
+
+let eval_term_generic
+  (eval_aux : IT.t -> (IT.t, string) result)
+  (prog5 : unit Mucore.mu_file)
+  (it : IT.t)
+  : (IT.t, string) result
+  =
+  let ( let@ ) = Result.bind in
+  let return = Result.ok in
+  let open IT in
+  let (IT (t_, bt, here)) = it in
+  let eval_num_binop = eval_num_binop eval_aux bt here in
+  match t_ with
+  | Const _ | Nil _ | MapDef _ -> return it
+  | Sym x -> Error (Sym.pp_string x ^ " is free")
+  (* Unary ops *)
+  | Unop (Not, it') ->
+    let@ it' = eval_aux it' in
+    (match it' with
+     | IT (Const (Bool b), _, _) -> return @@ bool_ (not b) here
+     | _ -> Error ("Not constant boolean (" ^ __LOC__ ^ ")"))
+  | Unop (Negate, it') ->
+    let@ it' = eval_aux it' in
+    (match it' with
+     | IT (Const (Z n), _, _) -> return @@ num_lit_ (Z.neg n) bt here
+     | IT (Const (Bits ((Signed, _), n)), _, _) ->
+       return @@ num_lit_ (BT.normalise_to_range_bt bt (Z.neg n)) bt here
+     | IT (Const (Bits ((Unsigned, _), _)), _, _) ->
+       Error
+         ("Can't negate unsigned integer at "
+          ^ Locations.simple_location here
+          ^ " ("
+          ^ __LOC__
+          ^ ")")
+     | _ -> Error ("Not constant integer or bitvector (" ^ __LOC__ ^ ")"))
+  | Unop (BW_CLZ_NoSMT, it') ->
+    let@ it' = eval_aux it' in
+    (match it' with
+     | IT (Const (Bits ((sgn, bits), n)), bt, _) ->
+       let open Int64 in
+       let reverse_bits (n : Z.t) : Z.t =
+         let rec aux (bits : int) (n : int64) (acc : int64) : int64 =
+           if bits = 0 then
+             acc
+           else (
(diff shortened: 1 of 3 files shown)


dc-mak:
It would be useful if you could explain _why_ as well as what. I'm off on holiday for until 12th, so I'll let Christopher handle this.

backend/cn/lib/eval.ml
```diff
@@ -0,0 +1,606 @@
...
+  let min, max = BT.bits_range (sgn, sz) in
+  match sgn with
+  | Unsigned ->
+    Ok
+      (if Z.gt n max then
+         Z.rem n (Z.add max Z.one)
+       else if Z.lt n Z.zero then
+         Z.add n max
+       else
+         n)
```
cp526: If `n` can go arbitrarily far out of the (min,max) range, then this has to be recursive. Maybe there's a stronger invariant that avoids that, however.

backend/cn/lib/eval.ml
```diff
@@ -0,0 +1,606 @@
...
+    ->
+    let@ () = check_bits_bt (sgn1, sz1) (sgn2, sz2) here in
+    (match sgn1 with
+     | Unsigned ->
+       let@ res = handle_bounds (sgn1, sz1) (f n1 n2) in
+       return @@ IT.num_lit_ res bt here
+     | Signed ->
+       let@ res = handle_bounds (sgn1, sz1) (f n1 n2) in
+       return @@ IT.num_lit_ res bt here)
+  | _ -> return @@ IT.arith_binop op (it1, it2) here
```
cp526: Integer operations in the specification language shouldn't generate over/underflow errors: all operations are total functions (as in the SMT language).

backend/cn/lib/eval.ml
```diff
@@ -0,0 +1,606 @@
...
+  : (IT.t, string) result
+  =
+  let ( let@ ) = Result.bind in
+  let return = Result.ok in
+  let open IT in
+  let (IT (t_, bt, here)) = it in
+  let eval_num_binop = eval_num_binop eval_aux bt here in
+  match t_ with
+  | Const _ -> return it
+  | Sym x -> Result.error (Sym.pp_string x ^ " is free")
```
cp526: If we want to be able to use `eval` to simplify quantified constraints or assertions, then it can't crash on free variables.

backend/cn/lib/eval.ml
```diff
@@ -0,0 +1,606 @@
...
+    let@ it' = eval_aux it' in
+    (match it' with
+     | IT (Const (Bool b), _, _) -> return @@ bool_ (not b) here
+     | _ -> return @@ not_ it' here)
+  | Unop (Negate, it') ->
+    let@ it' = eval_aux it' in
+    (match it' with
+     | IT (Const (Z n), _, _) -> return @@ num_lit_ (Z.neg n) bt here
+     | IT (Const (Bits ((Signed, sz), n)), _, _) ->
+       let@ res = handle_bounds (Signed, sz) (Z.neg n) in
```
cp526: same here as above: negation shouldn't crash if out of bounds.

backend/cn/lib/eval.ml
```diff
@@ -0,0 +1,606 @@
...
+       return @@ num_lit_ res bt here
+     | IT (Const (Bits ((Unsigned, _), _)), _, _) ->
+       Result.error ("Can't negate unsigned integer at " ^ Locations.simple_location here)
+     | _ -> return @@ negate it' here)
+  | Unop (BW_CLZ_NoSMT, _) -> failwith "todo: BW_CLZ_NoSMT"
+  | Unop (BW_CTZ_NoSMT, it') ->
+    let@ it' = eval_aux it' in
+    (match it' with
+     | IT (Const (Bits ((_, sz), n)), _, _) ->
+       let res = Z.trailing_zeros n in
```
cp526, line 115: Have to make sure Z's `trailing_zeros` agrees with how we've defined the SMT solver mapping in `solver.ml`.

backend/cn/lib/eval.ml
```diff
@@ -0,0 +1,606 @@
...
+  | Unop (BW_FFS_NoSMT, it') ->
+    let@ it' = eval_aux it' in
+    (match it' with
+     | IT (Const (Bits (_, n)), _, _) ->
+       if Z.equal n Z.zero then
+         return @@ num_lit_ Z.zero bt here
+       else
+         let@ it' = return @@ arith_unop BW_CTZ_NoSMT it' here in
+         eval_aux (add_ (it', num_lit_ Z.one bt here) here)
+     | _ -> return @@ arith_unop BW_FFS_NoSMT it' here)
```
cp526: as above: has to be aligned with solver.ml

backend/cn/lib/eval.ml
ZippeyKeys12: It doesn't crash, it returns `Error`, which `partial_eval` accounts for (and is what would be used for simplification).

backend/cn/lib/eval.ml
```diff
@@ -0,0 +1,607 @@
...
+          ^ " ("
+          ^ __LOC__
+          ^ ")")
+     | _ -> Error ("Not constant integer or bitvector (" ^ __LOC__ ^ ")"))
+  | Unop (BW_CLZ_NoSMT, it') ->
+    let@ it' = eval_aux it' in
+    (match it' with
+     | IT (Const (Bits ((sgn, bits), n)), bt, _) ->
+       let open Int64 in
+       let reverse_bits (n : Z.t) : Z.t =
```
cp526, line 93: I'm not sure this does the right thing, at least for bitvector sizes larger than 64?

cp526:
Aside from the inline comments, three general points:

- Is it correct that terms with free variables, such as `x == x` or `x + 0` will never be evaluated, even when they could be simplified? This is just to understand how this relates to the current term simplifier: as discussed previously, we probably eventually want to converge on just one.

- Is the (AIUI possibly diverging) lazy evaluation ever used?

- I'm a bit uneasy about the bitvector operations. AIUI, aside from some hand-coded unary operations, the strategy for binary operations is to simply compute in Z and then use the normalisation function to bring the result into the range of the bitvector type. But do we know whether that gives us the right semantics (i.e. the one given by the mapping to SMT)? For instance, the implementation of division seems to ignore the sign of the bitvector type, and there's probably various subtleties to get right in general. 

  How fast does the partial evaluator have to be? We could send bitvectors off to the SMT solver for evaluation (perhaps not as slow as solving), via the same `solver.ml` pipeline, to make sure the semantics is good.

ZippeyKeys12:
> * Is it correct that terms with free variables, such as `x == x` or `x + 0` will never be evaluated, even when they could be simplified? This is just to understand how this relates to the current term simplifier: as discussed previously, we probably eventually want to converge on just one.

Yes, it won't be evaluated. The goal here is just to do evaluation, which should be fairly simple/clear. For a term simplifier, you'd just call `eval` in it, similar to what `partial_eval` does.

> * Is the (AIUI possibly diverging) lazy evaluation ever used?

The lazy evaluation is the one I plan to use, and shouldn't diverge. Partial evaluation could, which is why we special-case it to fully evaluate arguments before application. It should also simplify more than the strict version.

> * I'm a bit uneasy about the bitvector operations. AIUI, aside from some hand-coded unary operations, the strategy for binary operations is to simply compute in Z and then use the normalisation function to bring the result into the range of the bitvector type. But do we know whether that gives us the right semantics (i.e. the one given by the mapping to SMT)? For instance, the implementation of division seems to ignore the sign of the bitvector type, and there's probably various subtleties to get right in general.

I removed the QCheck/OUnit tests I used to check them, to simplify the PR, but I can add them back and see if I missed any operations.

>   How fast does the partial evaluator have to be? We could send bitvectors off to the SMT solver for evaluation (perhaps not as slow as solving), via the same `solver.ml` pipeline, to make sure the semantics is good.

This is what the QCheck/OUnit tests were doing, and then seeing if `eval` and the SMT solver agreed. This also partly why the strict evaluation exists, to be easier to think about and to test the lazy evaluation against it.

cp526:
> > * Is it correct that terms with free variables, such as `x == x` or `x + 0` will never be evaluated, even when they could be simplified? This is just to understand how this relates to the current term simplifier: as discussed previously, we probably eventually want to converge on just one.
> 
> Yes, it won't be evaluated. The goal here is just to do evaluation, which should be fairly simple/clear. For a term simplifier, you'd just call `eval` in it, similar to what `partial_eval` does.
> 
> > * Is the (AIUI possibly diverging) lazy evaluation ever used?
> 
> The lazy evaluation is the one I plan to use, and shouldn't diverge. Partial evaluation could, which is why we special-case it to fully evaluate arguments before application. It should also simplify more than the strict version.
> 
> > * I'm a bit uneasy about the bitvector operations. AIUI, aside from some hand-coded unary operations, the strategy for binary operations is to simply compute in Z and then use the normalisation function to bring the result into the range of the bitvector type. But do we know whether that gives us the right semantics (i.e. the one given by the mapping to SMT)? For instance, the implementation of division seems to ignore the sign of the bitvector type, and there's probably various subtleties to get right in general.
> 
> I removed the QCheck/OUnit tests I used to check them, to simplify the PR, but I can add them back and see if I missed any operations.

Tests are good, but really the question is what the correctness argument is wrt the SMT mapping. 

For instance, in the SMT mapping CN Div turns into either an SMT `bv_sdiv` or `bv_udiv`, depending on the sign of the CN bitvector type. I don't see anything corresponding to that here. Does the partial evaluator's Z.div + normalisation behave like bv_sdiv and bv_udiv, respectively?

> 
> > How fast does the partial evaluator have to be? We could send bitvectors off to the SMT solver for evaluation (perhaps not as slow as solving), via the same `solver.ml` pipeline, to make sure the semantics is good.
> 
> This is what the QCheck/OUnit tests were doing, and then seeing if `eval` and the SMT solver agreed. This also partly why the strict evaluation exists, to be easier to think about and to test the lazy evaluation against it.

... Given the question about the relationship between the SMT semantics and the partial evaluator's implementation, would the evaluator be fast enough if we replaced its bitvector logic with a call into the SMT solver's evaluator (i.e. whenever a unary or binary operation is applied to a bitvector constant, simply asking the SMT solver to evaluate it for us)?

cp526:
@ZippeyKeys12 Any updates, what's the state of this?

Issue #68, opened 2018-07-03
Wrong printf arguments


victorgomes:
```c
#include <stdio.h>
int main(void){
	printf("%d\n", (long long int)(7));
}
```
I actually don't know if this is indeed UB, kcc claims that it is.

PR #459, opened 2024-08-02
Translate FunctionNoParams as a function without parameters.

### LINKED ISSUE #438: [CN] feature: allow mention of function pointers with empty parameter lists
```
struct foo {
    int (*bar)();
};
```
```
% cn verify empty_param.c 
empty_param.c:1:1: error: C-type signed int ()*
struct foo {
~~~~~~~^~~~~ 
```

This is holding up TA2 work. We *don't* need to be able to call these functions at this time. I expect calling them will require an ownership that includes an actual type for the pointer.

yav:
Fixes #438.

When this is merged, we should also merge PR #68 in the tutorial repo to fix the tests.

### DIFF
--- a/backend/cn/lib/sctypes.ml
+++ b/backend/cn/lib/sctypes.ml
@@ -171,7 +171,14 @@ let rec of_ctype (Ctype.Ctype (_, ct_)) =
     in
     let@ ret_ct = of_ctype ret_ct in
     return (Function ((ret_q, ret_ct), args, variadic))
-  | Ctype.FunctionNoParams _ -> fail
+
+  (* The meaning of a function with no parameters depends on the version of C,
+     but here we treat it as if it really has no parameter (i.e., like `void`)
+     which is what more recent versions of the standard do. *)
+  | Ctype.FunctionNoParams (ret_q, ret_ct) ->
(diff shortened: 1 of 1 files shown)


cp526:
Kayvan says there's an issue in Cerberus' frontend: Cerberus should've handled FunctionNoParams – which really is different from a function with empty parameter list – before this even gets to CN. In the absence of that, this is a good workaround.

backend/cn/lib/sctypes.ml
```diff
@@ -171,7 +171,14 @@ let rec of_ctype (Ctype.Ctype (_, ct_)) =
     in
     let@ ret_ct = of_ctype ret_ct in
     return (Function ((ret_q, ret_ct), args, variadic))
-  | Ctype.FunctionNoParams _ -> fail
+
+  (* The meaning of a function with no parameters depends on the version of C,
+     but here we treat it as if it really has no parameter (i.e., like `void`)
+     which is what more recent versions of the standard do. *)
```
cp526, line 177: (As below, this is ok as a workaround, but FunctionNoParams has to be treated specially, in the Cerberus frontend before CN.)

PR #461, opened 2024-08-04
Convert crash on binop to controlled error

septract:
Partially addresses #233

### DIFF
--- a/backend/cn/lib/check.ml
+++ b/backend/cn/lib/check.ml
@@ -737,7 +737,15 @@ let rec check_pexpr (pe : BT.t mu_pexpr) (k : IT.t -> unit m) : unit m =
            k x))
      | M_PEbounded_binop (M_Bound_Except act, iop, pe1, pe2) ->
        let@ () = WellTyped.WCT.is_ct act.loc act.ct in
-       let ity = match act.ct with Integer ity -> ity | _ -> assert false in
+       let@ ity =
+         match act.ct with
+         | Integer ity -> return ity
+         | _ ->
+           fail (fun _ ->
+             { loc;
+               msg = Generic !^"M_PEbounded_binop: unsupported on non-integer types"
+             })
+       in
        let@ () = ensure_base_type loc ~expect (Memory.bt_of_sct act.ct) in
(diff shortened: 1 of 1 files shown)


septract:
@cp526 I guessed at an error message, but perhaps you can propose something better?

backend/cn/lib/check.ml
```diff
@@ -693,10 +693,15 @@ let rec check_pexpr (pe : BT.t mu_pexpr) (k : IT.t -> unit m) : unit m =
...
-       assert (
+       let@ () =
          match act.ct with
-         | Integer ity when Sctypes.is_unsigned_integer_type ity -> true
-         | _ -> false);
+         | Integer ity when Sctypes.is_unsigned_integer_type ity -> return ()
+         | _ ->
+           fail (fun _ ->
+             { loc;
+               msg = Generic !^"M_PEbounded_binop: unsupported on non-integer types"
```
cp526, line 702: This should probably mention that this is unsupported on anything other than *unsigned* integers.

cp526:
... Although I think these might really be proper assertion failures.

septract:
What does it mean for this to be a proper assertion failure? I would assume these should only occur in anomalous states, ie states that should be prevented by some other CN code

cp526:
I suspect we're in that kind of situation here: something in CN+Cerberus's earlier logic should prevent an application of `M_PEbounded_binop (M_Bound_Wrap, ...)` when the C-type is a signed integer (wrapping behaviour typically only happens for unsigned types).

Similarly, the M_PEbounded_binop below should only happen on integer types.

septract:
Huh, okay - so is the conclusion that the crash in #233 is actually a bug in CN, rather than an unsupported feature? That seems more difficult to resolve, & probably needs you to take a deeper look @cp526

cp526:
Yes, I think so; I see no reason why CN shouldn't handle the example from #233 .

septract:
Okay, I'm closing this PR - looks like someone should look at the root cause of #233 (I'll take a look but I suspect it's too deep in the stack for my level of understanding)

Issue #412, opened 2024-07-23
[CN] Capture all output (not just return codes) for CI in a convenient to update way


dc-mak:
We don't want unnecessary duplication of C files, since any changes to syntax or semantics would need to be duplicated across multiple tests.

Perhaps if/when CN is a library, we can use some [expect tests](https://github.com/janestreet/ppx_expect) (including for tutorial. This would also be useful for checking the behaviour/output of tutorials once those are integrated. 

A small example I made a while back is here: https://github.com/dc-mak/NumLin/tree/master/test

dc-mak:
# Update

With the additions of 
* [runtime testing](https://github.com/rems-project/cerberus/commit/2702dbaed71f6695dc95a88b7543501ce1d8701c)
* [test generation](https://github.com/rems-project/cerberus/commit/c77d64541983966a50cc00e349b6627756d18024)
* and [multiple errors](https://github.com/rems-project/cerberus/pull/474)
simple return codes are no longer sufficient to capture the behaviour of CN in various modes.

I propose that, _in addition to return codes_, for a given files we have the ability to specify the expected output in a `filename.expected` if we want to to use a script [like this](https://github.com/rems-project/c-tree-carver/blob/main/cpp/test/run_test.py) or use [expect tests](https://github.com/janestreet/ppx_expect) as stated above. A key feature of both is that tests can be auto-promoted when behaviour changes en masse

Issue #299, opened 2024-06-04
[CN] Make return codes mean something consistently


dc-mak:
Return error code 2 is currently overloaded for "cerberus error" and "cn unimplemented" and those are two very different things. 

https://github.com/rems-project/cerberus/commit/3b0bda322d0c2e2488def31db94a4f88392d1ea2

This is a nice-to-have feature which is easy to do, but would require some work to recategorise all the CI examples.

bcpierce00:
Seems worth doing sooner rather than later...

septract:
I may look at this next, unless someone else is already addressing it. 

We may want to move towards the Rust compiler practice, where each distinct error type has a distinct error code, broken down in quite fine detail. Rust then has a [very large table](https://doc.rust-lang.org/error_codes/error-index.html) with all of the possible errors, along with possible remedies / advice for each error. When the compiler throws an error, it outputs the URL pointing to the suggested remedy. 

For now, I think it'd be useful to just wrap all calls to `exit` (and `failwith`) in some consistent abstraction, and somewhere write down the different types of errors we expect to see.

cp526:
Sounds good. `typeErrors.ml` defines an AST of error messages, so assigning a unique number to each shouldn't be hard.
We do have quite a few different types of errors, though, and there may be some that are unused (now or in the future) and need to be cleaned up.

septract:
I propose we use the following exit code scheme: 

- 0: success 
- 1: 'regular' errors caused by normal program states (eg. type error, parse error). In this case, we emit an error code corresponding to the particular error. 
- 2: uncaught exceptions (might not happen, see below?) 
- 125: internal errors / crashes / bugs. This already seems to be the behavior of OCaml. 

The documentation for OCaml `exit` is [here](https://ocaml.org/manual/5.2/api/Stdlib.html#1_Programtermination). It looks like we should avoid calling `exit 2` - according to the documentation, this is specifically used to denote an uncaught exception. (Though I'm a bit confused why the documentation claims that uncaught exceptions cause error 2, but in fact, crashes result in error 125. So maybe the documentation is wrong?) 

The `man` page that results from `cn --help` is wrong about how exit codes are used. It just uses boilerplate derived from the `Cmdliner` module. To correct it, we need to set a `Cmdliner` module parameter - details [here](https://erratique.ch/software/cmdliner/doc/Cmdliner/Cmd/index.html#type-info). 

[...]

PR #474, opened 2024-08-07
[CN] Report multiple verification errors, when possible

### LINKED ISSUE #287: [CN] `cn` can't report rich errors in multiple unrelated functions
Consider the following code:
```c
int add1(int x)
/*@
requires x == 0i32;
ensures return == 2i32;
@*/
{
    return x + 1;
}

int add3(int x)
/*@
requires x == 0i32;
ensures return == 2i32;
@*/
{
    return x + 3;
}
```

`cn` correctly reports that `add3`'s spec doesn't hold, but doesn't say anything about `add1`, whose spec is also wrong:
```
> cn sample/add1.c
[1/2]: add3
sample/add1.c:16:5: error: Unprovable constraint
    return x + 3;
    ^~~~~~~~~~~~~ 
Constraint from sample/add1.c:13:9:
ensures return == 2i32;
        ^~~~~~~~~~~~~~~
```

 `cn --batch` will report that both functions "fail", but will not report why:
```
> cn --batch sample/add1.c
[1/2]: add3 -- fail
[2/2]: add1 -- fail
summary:
0 pass, 2 fail
```

It would be more user-friendly to be able to report as many errors as richly as possible - in this case, success would look like `cn` reporting both erroneous specs, with their locations, in a single invocation.

(I thought there might already be an issue for this behavior, but I couldn't find one. If there is one, I'm happy to deduplicate and direct conversation there.)

samcowger:
Fixes #287, according to the discussion in that issue. In particular:
- CN now displays verification errors for multiple functions by default, if the functions under analysis pass their wellformedness checks.
- A new flag, `--fail-fast`, allows reversion to the preexisting single-error behavior.
- A new flag, `--quiet`, causes CN to only report pass/fail outcomes, like how `--batch` currently behaves.
- The `--batch` flag has been removed, supplanted by `--quiet`.
- The `--state-file` flag has been removed, with the hope that `--output-dir` provides users sufficient control over output.
- CN now checks functions in declaration order.

This ended up being a more substantial change than I'd hoped to make, in part because CN's assumption of only handling a single error was more baked-in than I'd anticipated, and in part because I took the liberty of doing some light refactoring and documentation in places where I needed to make changes anyway.

### DIFF
--- a/backend/cn/bin/main.ml
+++ b/backend/cn/bin/main.ml
@@ -188,12 +188,35 @@ let with_well_formedness_check
     Printexc.raise_with_backtrace exc (Printexc.get_raw_backtrace ())
 
 
-let handle_type_error ~json ~state_file ~output_dir e =
+(** Report an error on [stderr] in an appropriate format: JSON if [json] is
+    true, or human-readable if not. *)
+let report_type_error
+  ~(json : bool)
+  ?(state_file : string option)
+  ?(output_dir : string option)
+  (error : TypeErrors.t)
+  : unit
+  =
   if json then
-    TypeErrors.report_json ?state_file ?output_dir e
+    TypeErrors.report_json ?state_file ?output_dir error
   else
-    TypeErrors.report_pretty ?state_file ?output_dir e;
-  match e.msg with TypeErrors.Unsupported _ -> exit 2 | _ -> exit 1
+    TypeErrors.report_pretty ?state_file ?output_dir error
+
+
+(** Exit with a code appropriate to the provided error. *)
+let exit_on_type_error (error : TypeErrors.t) : 'a =
+  match error.msg with TypeErrors.Unsupported _ -> exit 2 | _ -> exit 1
+
+
+(** Report the provided error, then exit. *)
+let handle_type_error
+  ~(json : bool)
+  ?(state_file : string option)
+  ?(output_dir : string option)
+  (error : TypeErrors.t)
+  =
+  report_type_error ~json ?state_file ?output_dir error;
+  exit_on_type_error error
 
 
 let well_formed
@@ -222,7 +245,7 @@ let well_formed
     ~use_peval
     ~no_inherit_loc
     ~magic_comment_char_dollar
-    ~handle_error:(handle_type_error ~json ~state_file ~output_dir)
+    ~handle_error:(handle_type_error ~json ?state_file ?output_dir)
     ~f:(fun ~prog5:_ ~ail_prog:_ ~statement_locs:_ ~paused:_ -> Resultat.return ())
 
 
@@ -303,7 +326,7 @@ let verify
     ~use_peval
     ~no_inherit_loc
     ~magic_comment_char_dollar (* Callbacks *)
-    ~handle_error:(handle_type_error ~json ~state_file ~output_dir)
+    ~handle_error:(handle_type_error ~json ?state_file ?output_dir)
     ~f:(fun ~prog5 ~ail_prog ~statement_locs ~paused ->
       match output_decorated with
       | None ->
@@ -524,8 +547,6 @@ module Verify_flags = struct
   let fail_fast =
     let doc = "Abort immediately after encountering a verification error" in
     Arg.(value & flag & info [ "fail-fast" ] ~doc)
-
-
 
 
   let slow_smt_threshold =
--- a/backend/cn/bin/main.ml
+++ b/backend/cn/bin/main.ml
@@ -190,17 +190,13 @@ let with_well_formedness_check
 
 (** Report an error on [stderr] in an appropriate format: JSON if [json] is
     true, or human-readable if not. *)
-let report_type_error
-  ~(json : bool)
-  ?(state_file : string option)
-  ?(output_dir : string option)
-  (error : TypeErrors.t)
+let report_type_error ~(json : bool) ?(output_dir : string option) (error : TypeErrors.t)
   : unit
   =
   if json then
-    TypeErrors.report_json ?state_file ?output_dir error
+    TypeErrors.report_json ?output_dir error
   else
-    TypeErrors.report_pretty ?state_file ?output_dir error
+    TypeErrors.report_pretty ?output_dir error
 
 
 (** Exit with a code appropriate to the provided error. *)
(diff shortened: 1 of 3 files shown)


cp526:
This looks very good to me. 

But it seems that a number of CI tests are now failing. I wonder whether those are all tests that are expected to fail with html output but are crashing while producing that (leading to a 125 exit code)?

samcowger:
> This looks very good to me.
> 
> But it seems that a number of CI tests are now failing. I wonder whether those are all tests that are expected to fail with html output but are crashing while producing that (leading to a 125 exit code)?

Indeed they are, and I see why. I had made a change to the state file naming scheme, intending to include the _name_ of the file in which the error originated, but I accidentally included the _path_ to that file, so the generated name was invalid. The fix for that is simple.

I see another issue, though, which is that my changes don't preserve the existing error code semantics, so CN now exits with 0 even in the presence of verification errors. I can think of a few solutions, but I'd appreciate your opinion on which one seems most appropriate:
- Generate the exit code according to the first error encountered
  - Or, the last error encountered
- Always exit with a particular value in the presence of more than one error
- Create a severity-based ordering of errors, and generate the exit code for the most severe error encountered
- ...other ideas also welcomed!

cp526:
> > This looks very good to me.
> > But it seems that a number of CI tests are now failing. I wonder whether those are all tests that are expected to fail with html output but are crashing while producing that (leading to a 125 exit code)?
> 
> Indeed they are, and I see why. I had made a change to the state file naming scheme, intending to include the _name_ of the file in which the error originated, but I accidentally included the _path_ to that file, so the generated name was invalid. The fix for that is simple.

Great!

> 
> I see another issue, though, which is that my changes don't preserve the existing error code semantics, so CN now exits with 0 even in the presence of verification errors. I can think of a few solutions, but I'd appreciate your opinion on which one seems most appropriate:
> 
> * Generate the exit code according to the first error encountered
>   
>   * Or, the last error encountered
> * Always exit with a particular value in the presence of more than one error
> * Create a severity-based ordering of errors, and generate the exit code for the most severe error encountered
> * ...other ideas also welcomed!

No strong opinion, all those sound reasonable to me, with perhaps a slight preference for option (3), specially signalling multiple errors (and not depending on the order of functions as in options 1 or 2). 

As I understand it, gcc has some kind of "worst-error-code" option ( https://gcc.gnu.org/onlinedocs/gcc-8.2.0/gcc/Overall-Options.html#index-pass-exit-codes ). Though it's not clear to me that we have a meaningful severity order: what's worse, CN crashed on the input, or it successfully analysed the program but found it has undefined behaviour? 

We've separately discussed, I think, potentially moving to a scheme with fine-grained exit codes, one per kind of type error.

Another thing to consider: our main use of exit codes is the CI script. Should we run that with fail-fast mode? If we wanted to run CN without fail-fast (but I'm not sure we do) what's best for that?

samcowger:
I think that (3) is a reasonable choice - we can revisit it if we do end up moving towards encoding more granular information in error codes.

> Another thing to consider: our main use of exit codes is the CI script. Should we run that with fail-fast mode? If we wanted to run CN without fail-fast (but I'm not sure we do) what's best for that?

That's a good question - I lean towards not using `--fail-fast` in CI. Running exit-code-based CI with `--fail-fast` will alert us only if a file goes from passing verification to failing it, or vice versa. This is the semantics of CI now, absent this PR. Running without `--fail-fast` will still alert us in that situation, but may also alert us to changes in the number of verification failures in a file, which is valuable information for CI to expose. (I say "may" because we wouldn't learn if a function went from having e.g. two failing functions to having three - not good, but also not something we handle now, nor can we handle it in general with exit-code-based CI.)

cp526:
Sounds good to me.

samcowger:
Aha - now that I've introduced the new exit code, CI has caught that `tests/cn/reverse.error.c` triggers multiple errors:
```
> cn verify tests/cn/reverse.error.c
[1/2]: reverse -- fail
[2/2]: main -- fail
tests/cn/reverse.error.c:106:7: error: Unprovable constraint
      return cur;
      ^~~~~~~~~~~ 
Constraint from tests/cn/reverse.error.c:89:14:
             L_ == rev(L);
             ^~~~~~~~~~~~~ 
State file: file:///var/folders/c0/2x68zj1d0cl01nczx_nl283r0000gp/T/state_reverse.error.c_eb30dc.html
tests/cn/reverse.error.c:124:3: error: Missing resource for de-allocating
  return 0;
  ^~~~~~~~~ 
Resource needed: Block<struct node>(&n3)
  which requires: Block<signed int>(member_shift<node>(&n3, head))
      other location (Cn__ResourceInference.General.predicate_request)  (arg head)
State file: file:///var/folders/c0/2x68zj1d0cl01nczx_nl283r0000gp/T/state_reverse.error.c_d363da.html
```

It seems like the error CN is meant to find is the one in the function `reverse`, because only `reverse` differs semantically between `reverse.error.c` and its sibling `reverse.c`. I'll go ahead and add a `trusted` pragma to `main`, following the example set by https://github.com/rems-project/cerberus/commit/eb3e99060ff47288e1b562d4d047a70bb0debac7.

samcowger:
Looks like more tests fail for similar reasons, though unlike `reverse.error.c`, I think at least some of them actually have multiple semantic failures. This is exactly what I'd hoped CI would discover, but in the future, not in the present! I think the proper fix would involve triaging those failures to determine which ones we really expect to produce multiple errors and updating CI and/or the file tree organization to make sure CI recognizes that these tests are expected to fail with this new code.

This is enough design and implementation work that I would be inclined to defer it to an issue/a future PR, rather than wrap it up in this PR, especially since there would be a lot of overlap between that work and any future efforts to add granularity to exit codes. If we do defer, I think the most appropriate immediate fix would be to add the new multi-error exit code to CI so that it's recognized as an error like any other, which should preserve the existing testing semantics.

cp526:
> Looks like more tests fail for similar reasons, though unlike `reverse.error.c`, I think at least some of them actually have multiple semantic failures. This is exactly what I'd hoped CI would discover, but in the future, not in the present! I think the proper fix would involve triaging those failures to determine which ones we really expect to produce multiple errors and updating CI and/or the file tree organization to make sure CI recognizes that these tests are expected to fail with this new code.
> 
> This is enough design and implementation work that I would be inclined to defer it to an issue/a future PR, rather than wrap it up in this PR, especially since there would be a lot of overlap between that work and any future efforts to add granularity to exit codes. If we do defer, I think the most appropriate immediate fix would be to add the new multi-error exit code to CI so that it's recognized as an error like any other, which should preserve the existing testing semantics.

Sounds good to me.

dc-mak:
Potentially relevant to consider either before or soon after this PR: https://github.com/rems-project/cerberus/issues/412

samcowger:
After reviewing #299, I'm actually going to backtrack on adding a new exit code to signal multiple errors, and use one of the two existing exit codes instead. I've created an ad-hoc severity-based ordering, so that any "Unsupported" errors will dominate any other errors in determining which code to exit with. I think this mimics existing semantics well enough, and even if it's a little kludgy, it'll be overhauled when addressing #299.

My earlier point about capturing verification behavior changes in CI is subsumed by #412, I believe, so I will not make a new issue about it after all.

I think this PR is now ready to go (pending review/approval, of course).

cp526:
Looks good to me, thanks very much!

PR #476, opened 2024-08-08
[CI] unsilence CN output

samcowger:
This makes CI logs more verbose, but also much more useful for diagnostic purposes.

### DIFF
--- a/tests/run-cn.sh
+++ b/tests/run-cn.sh
@@ -35,8 +35,8 @@ function exits_with_code() {
   local file=$2
   local -a expected_exit_codes=$3
 
-  printf "[$file]... "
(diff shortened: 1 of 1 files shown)


samcowger:
If this change is deemed appropriate, it looks like there are a couple of places in `cn-tutorial` where we might also want to apply it:
- [`src/example-archive/check.sh`](https://github.com/rems-project/cn-tutorial/blob/1117099d2f1c7285eeee8b1fa270f196049a563d/src/example-archive/check.sh#L48)
- [`runtime-test.sh`](https://github.com/rems-project/cn-tutorial/blob/b762b022f6d8bc8fb440464c01c73bacc37ebc78/runtime-test.sh#L25) (on the `cn-runtime-testing` branch)

PR #477, opened 2024-08-08
[CN] Add support for shift-right

### LINKED ISSUE #464: [CN] Confusing error message for shift right
For this minimal function:
```
int shift(int n){
        return n>>2;
}
```

Running CN: >cn verify shr.c
shr.c:2:9: error: PEcall to impl not inlined: SHR_signed_negative
        return n>>2;

Does this mean SHR is implemented or not? Thanks!

yav:
This fixes #464.   The issue is about shifting right on negative numbers, which is implementation defined behavior.
The PR also adds some comments to `mu_core.mli`.

The update adds another special case to `translation.lem` to avoid turn a shift into an exponentiation, because reasoning about exponentiation is hard.

We also add a new case in `core-to-mu-core` to handle the case when we are shifting a negative number.
The case has a flag that controls if we should report this as an error or translate to an arithmetic shift.
At the moment the flag is hard-coded to the latter.

### DIFF
--- a/backend/cn/lib/check.ml
+++ b/backend/cn/lib/check.ml
@@ -544,8 +544,9 @@ let rec check_pexpr (pe : BT.t mu_pexpr) (k : IT.t -> unit m) : unit m =
              { loc; msg = Mismatch { has = BT.pp ty; expect = !^"comparable type" } })
        in
        let not_yet x =
-           Pp.debug 1 (lazy (Pp.item "not yet restored" (Pp_mucore_ast.pp_pexpr orig_pe)));
-           failwith ("todo: " ^ x) in
+         Pp.debug 1 (lazy (Pp.item "not yet restored" (Pp_mucore_ast.pp_pexpr orig_pe)));
+         failwith ("todo: " ^ x)
+       in
        (match op with
         | OpDiv ->
           let@ () = WellTyped.ensure_base_type loc ~expect (bt_of_pexpr pe1) in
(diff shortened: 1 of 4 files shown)


yav:
I just found `ocaml_implementation.ml`, so I realize that instead of catching `CF.Implementation.SHR_signed_negative` in `core_to_mucore` one should probably use this framework for the implementation specific behavior.

Since we are trying to restrict changes to the CN portion of the code base, and this is a blocking issue, I think the ucrrent implementation makes sense.  On Cerberus implements this correctly, the code in `core_to_mucore` should become unreachable and can be deleted in the longer run.

kmemarian:
in the diff to `translation.lem` why is it using `catch_exceptional_condition` instead of `wrapI`?

yav:
I used `catch_exceptional_condition` because this is in the signed case.   In this particular case nothing should really be overflowing, so I don't know that it matters, but it looks like this is what's done in the other operations, so perhaps it's better to be consistent.   Also, CN has an assert that `wrap` is only used with unsigned types, so I think we have to use `exception_condition`, unless we want to change *that* assert instead.

By the way, it looks like we have the same issue as #233 here as well, where the signedness check is done on the *unpromoted* type of the first argument.

kmemarian:
> I used `catch_exceptional_condition` because this is in the signed case. In this particular case nothing should really be overflowing, so I don't know that it matters, but it looks like this is what's done in the other operations, so perhaps it's better to be consistent. Also, CN has an assert that `wrap` is only used with unsigned types, so I think we have to use `exception_condition`, unless we want to change _that_ assert instead.

Ah ok, that makes sense.

> By the way, it looks like we have the same issue as #233 here as well, where the signedness check is done on the _unpromoted_ type of the first argument.

Yes I noticed that. I'll merge this PR and then commit the fixes you suggested to `<<` and `>>` to avoid conflicts here.

PR #479, opened 2024-08-09
Use model to simplify report

yav:
This PR does the following:
  * Adds a .mli file and some docs to the `Explain` and `Report` modules
  * Modifies `Explain` to do simplify things based on the current model. 
     - For constraints, we eliminate disjunctions, implications, and if-then-else, by eliminating the alternatives that are not viable in the model
     - For the resource hint, we eliminate constraints that are true in the model

### DIFF
--- a/backend/cn/lib/explain.ml
+++ b/backend/cn/lib/explain.ml
@@ -88,6 +88,34 @@ let subterms_without_bound_variables bindings =
     ITSet.empty
 
 
+(** Simplify a constraint in the context of a model. *)
+let simp_constraint eval lct =
+  let eval_to_bool it =
+    match eval it with Some (IT (Const (Bool b1), _, _)) -> Some b1 | _ -> None
+  in
+  let is b it = match eval_to_bool it with Some b1 -> Bool.equal b b1 | _ -> false in
+  let rec go (IT (term, bt, loc)) =
+    let mk x = IT (x, bt, loc) in
+    let ands xs = IT.and_ xs loc in
+    let go1 t = ands (go t) in
+    match term with
+    | Const (Bool true) -> []
+    | Binop (Or, lhs, rhs) when is false lhs -> go rhs
+    | Binop (Or, lhs, rhs) when is false rhs -> go lhs
+    | Binop (And, lhs, rhs) -> List.append (go lhs) (go rhs)
+    | Binop (Implies, lhs, rhs) ->
+      (match eval_to_bool lhs with
+       | Some b -> if b then go rhs else []
+       | None -> [ mk (Binop (Implies, go1 lhs, go1 rhs)) ])
+    | ITE (cond, ifT, ifF) ->
+      (match eval_to_bool cond with
+       | Some b -> if b then go ifT else go ifF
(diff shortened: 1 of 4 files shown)


cp526:
Thanks! That looks like it should be very useful. 

I wonder whether we might want to set it up so there's a switch in the top bar of the html that toggles between two views:
(1) Showing the state, including constraints and resources just as CN "knows" them, not specialised to the counterexample.
(2) The state specialised to the counterexample, including specialised versions of constraints and resources.

That could be quite easy to implement, by just always recording two versions of some of the components in `Report.report`, and letting the html switch between these.

The UI should (eventually) be useful for two purposes:

(a) For debugging a verification failure. Here I imagine (2) would often be really useful, to reduce the noise in what's displayed and to make it easier to understand the gap in the proof. One probably still wants to sometimes also see (1). For instance when the verification requires a new lemma, if that's an inductive fact, it might be useful to see (1) to work out how to generalise the current counterexample.

(b) A new CN user should be able to use the interactive UI to develop an intuition for how CN works: for instance by stepping through some (verified or unverified) example to see what CN "knows" at any point along the symbolic execution path. For this (1) is more useful than (2), because the point is not to understand a concrete failure.


Separately: it's probably useful to also be able to switch between showing specialised and non-specialised versions of the unproved constraint or missing resource causing the error (the `extras` component in `Explain`).

backend/cn/lib/explain.ml
```diff
@@ -88,6 +89,53 @@ let subterms_without_bound_variables bindings =
...
+      (match eval_to_bool lhs with
+       | Some b -> if b then go rhs else []
+       | None -> [ mk (Binop (Implies, go1 lhs, go1 rhs)) ])
+    | ITE (cond, ifT, ifF) ->
+      (match eval_to_bool cond with
+       | Some b -> if b then go ifT else go ifF
+       | None -> [ mk (ITE (go1 cond, go1 ifT, go1 ifF)) ])
+    | _ -> [ mk term ]
+  in
+  match lct with LC.Forall _ -> [ lct ] | LC.T ct -> List.map (fun x -> LC.T x) (go ct)
```
cp526, line 117: We might want to extend `go` to know about binders (and not try to evaluate those), so we can apply the evaluation to `Forall` and to iterated resources.

PeterSewell:
+1 for that switch

backend/cn/lib/report.mli
```diff
@@ -23,21 +19,36 @@ type where_report =
...
+    the first component is "interesting", the second is not. *)
 type state_report =
-  { where : where_report;
-    resources : Pp.document list * Pp.document list;
-    constraints : Pp.document list * Pp.document list;
-    terms : term_entry list * term_entry list
+  { where : where_report; (** Location information *)
+    resources : Pp.document list * Pp.document list; (** Resources *)
+    constraints : Pp.document list * Pp.document list; (** Constraints *)
+    terms : term_entry list * term_entry list (** Term values *)
```
dc-mak, line 34: Just for future reference, my preference is that comments should be there to explain _why_ rather than what and variable names should be self explanatory as far as possible. If the variable name would end up being too long, then a short name plus a comment (e.g. as in lines 39-43) is good.

Issue #490, opened 2024-08-13
[CN] Remove `list`, and remove / deprecate other unused keywords


septract:
**Problem:** 

* CN considers `list` a keyword, but doesn't use it for any current features. See the discussion [here](https://mattermost.galois.com/verse-ta1-ta2/pl/kxkcwbmm37rm7dbfehkf6mb5oe). We would like to free it up for use in examples. 
* CN considers `pack` / `unpack` keywords, but they are not currently hooked up. Using these keywords results in various unhelpful errors, but does not clearly say that these are unimplemented. 
* CN may also include other features that are dead / deprecated - e.g the VSCode plugin marks `tuple` as deprecated. 

**Proposed resolution:**

* Remove `list` as a keyword, and all downstream dead code that results. 
* Keep `pack` / `unpack` as reserved keywords, but on any use, emit an error saying they are unsupported. Remove all downstream dead code that results. 
* Deal with `tuple` and any other keywords by either removing or deprecating them. 

@bcpierce00 @cp526

bcpierce00:
+1

dc-mak:
Finished by https://github.com/rems-project/cerberus/pull/497

PR #497, opened 2024-08-14
Cleanup keywords - addresses #490 

septract:
Addresses #490 

This PR: 
* Breaks keywords into 3 categories - production, experimental, and unsupported
  - 'production' keywords behave as normal 
  - 'experimental' keywords raise a warning, and then continue verification 
  - 'unimplemented' keywords raise an error, but are reserved for future use 
* Renames all keywords in the 'experimental' group from `name` -> `cn_name` 
* Removes `bool` and `CN_bool` as synonyms for `boolean` 
* Updates the tests to follow the new syntax

The KW categories are: 
* 'experimental': `cn_list`, `cn_tuple`, `cn_map`, `cn_set`, `cn_have`, `cn_function` (not renamed), `cn_print`
* 'unimplemented': `pack`, `unpack` 
* 'production': everything else 

Input needed: 
* Are these categories correct? 
* Should we have a way to suppress the warning on experimental keywords? (eg. a `--experimental` flag)

### DIFF
--- a/ocaml_frontend/pprinters/pp_errors.ml
+++ b/ocaml_frontend/pprinters/pp_errors.ml
@@ -57,7 +57,7 @@ let string_of_cparser_cause = function
   | Cparser_mismatched_magic_comment ->
       "magic comment syntax at open but not close"
   | Cparser_experimental_keyword str -> 
-      "experimental keyword '" ^ str ^ "'. Use of experimental features is discouraged"
+      "experimental keyword '" ^ str ^ "' (use of experimental features is discouraged)"
   | Cparser_unimplemented_keyword str -> 
       "unimplemented keyword '" ^ str ^ "'"
 
--- a/ocaml_frontend/pprinters/pp_errors.ml
+++ b/ocaml_frontend/pprinters/pp_errors.ml
@@ -56,8 +56,10 @@ let string_of_cparser_cause = function
       "unterminated " ^ str ^ " comment"
   | Cparser_mismatched_magic_comment ->
       "magic comment syntax at open but not close"
-  | Cparser_deprecated_keyword str -> 
-      "keyword '" ^ str ^ "' is deprecated and should not be used"
+  | Cparser_experimental_keyword str -> 
+      "experimental keyword '" ^ str ^ "'. Use of experimental features is discouraged"
+  | Cparser_unimplemented_keyword str -> 
+      "unimplemented keyword '" ^ str ^ "'"
 
 
 let string_of_constraint_violation = function
--- a/ocaml_frontend/pprinters/pp_errors.ml
+++ b/ocaml_frontend/pprinters/pp_errors.ml
@@ -56,6 +56,8 @@ let string_of_cparser_cause = function
       "unterminated " ^ str ^ " comment"
   | Cparser_mismatched_magic_comment ->
       "magic comment syntax at open but not close"
+  | Cparser_deprecated_keyword str -> 
+      "keyword '" ^ str ^ "' is deprecated and should not be used"
 
 
 let string_of_constraint_violation = function
--- a/parsers/c/c_lexer.mll
+++ b/parsers/c/c_lexer.mll
@@ -168,32 +168,29 @@ let cn_lex_builder kw_list : (string, token) Hashtbl.t  =
   let add (key, builder) = Hashtbl.add cn_lex key builder in
   List.iter add kw_list; cn_lex
 
-let cn_lexicon: (string, token) Hashtbl.t = 
-  cn_lex_builder cn_keywords
-
-let cn_lexicon_experimental: (string, token) Hashtbl.t = 
-  cn_lex_builder cn_keywords_experimental
-
-let cn_lexicon_unimplemented: (string, token) Hashtbl.t = 
-  cn_lex_builder cn_keywords_unimplemented
-
+(* 
+Attempt to lex a CN keyword. These may be: 
+  * 'production' - well-supported and suitable for general use
+  * 'experimental' - functional in some cases but not recommended for general use 
+  * 'unimplemented' - non-functional, but the keyword is reserved  
+*)
 let cn_lex_keyword id start_pos = 
   (* Try to lex CN production keywords *)
-  try Hashtbl.find cn_lexicon id
+  try Hashtbl.find (cn_lex_builder cn_keywords) id
   with Not_found ->
     (* Try to lex CN experimental keywords - warn on success *)
     try 
-      let kw = Hashtbl.find cn_lexicon_experimental id in 
+      let kw = Hashtbl.find (cn_lex_builder cn_keywords_experimental) id in 
       prerr_endline       
         (Pp_errors.make_message
          (Cerb_location.point start_pos)
-         Errors.(CPARSER (Cparser_experimental_keyword id))
+         Errors.(CPARSER (Errors.Cparser_experimental_keyword id))
          Warning);
       kw 
     with Not_found ->
       try 
         (* Try to lex CN unimplemented keywords - throw an error on success *)
-        let _ = Hashtbl.find cn_lexicon_unimplemented id in 
+        let _ = Hashtbl.find (cn_lex_builder cn_keywords_unimplemented) id in 
         raise (Error (Errors.Cparser_unimplemented_keyword id))
       with Not_found ->
         UNAME id
--- a/parsers/c/c_lexer.mll
+++ b/parsers/c/c_lexer.mll
@@ -112,7 +112,6 @@ let cn_keywords: (string * Tokens.token) list = [
     "pointer"       , CN_POINTER;
     "alloc_id"      , CN_ALLOC_ID;
     "map"           , CN_MAP;
-    "list"          , CN_LIST;
     "tuple"         , CN_TUPLE;
     "set"           , CN_SET;
     "let"           , CN_LET;
--- a/parsers/c/c_lexer.mll
+++ b/parsers/c/c_lexer.mll
@@ -113,6 +113,7 @@ let cn_keywords: (string * Tokens.token) list = [
     "real"          , CN_REAL;
     "pointer"       , CN_POINTER;
     "alloc_id"      , CN_ALLOC_ID;
+    "map"           , CN_MAP;
     "let"           , CN_LET;
     "take"          , CN_TAKE;
     "Owned"         , CN_OWNED;
@@ -150,7 +151,6 @@ unsuitable for non-experts *)
 let cn_keywords_experimental: (string * Tokens.token) list = [
     "cn_list"       , CN_LIST; 
     "cn_tuple"      , CN_TUPLE;
-    "cn_map"        , CN_MAP;
     "cn_set"        , CN_SET;
     "cn_have"       , CN_HAVE;
     "cn_function"   , CN_FUNCTION;
@@ -188,12 +188,9 @@ let cn_lex_keyword id start_pos end_pos =
          Warning);
       kw 
     with Not_found ->
-      try 
-        (* Try to lex CN unimplemented keywords - throw an error on success *)
-        let _ = Hashtbl.find (cn_lex_builder cn_keywords_unimplemented) id in 
-        raise (Error (Errors.Cparser_unimplemented_keyword id))
-      with Not_found ->
-        UNAME id
+      (* Try to lex CN unimplemented keywords - throw an error on success *)
+      let _ = Hashtbl.find (cn_lex_builder cn_keywords_unimplemented) id in 
+      raise (Error (Errors.Cparser_unimplemented_keyword id))
 
 (* END CN *)
 
@@ -595,7 +592,10 @@ and initial flags = parse
   | ['A'-'Z']['0'-'9' 'A'-'Z' 'a'-'z' '_']* as id
       {
         if flags.inside_cn then
-          cn_lex_keyword id lexbuf.lex_start_p lexbuf.lex_curr_p  
+          try 
+            cn_lex_keyword id lexbuf.lex_start_p lexbuf.lex_curr_p  
+          with Not_found ->
+            UNAME id 
         else
           UNAME id
       }
@@ -605,7 +605,10 @@ and initial flags = parse
         Hashtbl.find lexicon id
       with Not_found ->
         if flags.inside_cn then
-          cn_lex_keyword id lexbuf.lex_start_p lexbuf.lex_curr_p 
+          try 
+            cn_lex_keyword id lexbuf.lex_start_p lexbuf.lex_curr_p 
+          with Not_found ->
+            LNAME id 
         else
           LNAME id
     }
--- a/parsers/c/c_lexer.mll
+++ b/parsers/c/c_lexer.mll
@@ -93,78 +93,74 @@ let lexicon: (string, token) Hashtbl.t =
 
 (* BEGIN CN *)
 
-(* CN 'production' keywords: well-supported and suitable for general use *)
-let cn_keywords: (string * Tokens.token) list = [
-    "good"          , CN_GOOD;
-    "boolean"       , CN_BOOL;
-    "integer"       , CN_INTEGER;
-    "u8"           , CN_BITS (`U,8);
-    "u16"           , CN_BITS (`U,16);
-    "u32"           , CN_BITS (`U,32);
-    "u64"           , CN_BITS (`U,64);
-    "u128"           , CN_BITS (`U,128);
-    "i8"           , CN_BITS (`I,8);
-    "i16"           , CN_BITS (`I,16);
-    "i32"           , CN_BITS (`I,32);
-    "i64"           , CN_BITS (`I,64);
-    "i128"           , CN_BITS (`I,128);
-    "real"          , CN_REAL;
-    "pointer"       , CN_POINTER;
-    "alloc_id"      , CN_ALLOC_ID;
-    "map"           , CN_MAP;
-    "let"           , CN_LET;
-    "take"          , CN_TAKE;
-    "Owned"         , CN_OWNED;
-    "Block"         , CN_BLOCK;
-    "each"          , CN_EACH;
-    "NULL"          , CN_NULL;
-    "true"          , CN_TRUE;
-    "false"         , CN_FALSE;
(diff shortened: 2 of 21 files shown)


septract:
Questions for @cp526:
* Does CN need three different keywords for boolean (`bool`, `boolean` and `CN_bool`). Could we standardize on `bool`, which is what the pretty-printer emits? 
* Is the `set` keyword used? Could we deprecate it? 
* Could we remove the `CN_list` type? What about the `List` constructor for the `BT` type? 
* For `tuple` / `pack` / `unpack` - I guess you want to keep the code implementing this functionality in place, and just disable parsing of the keywords?

cp526:
> Questions for @cp526:
> 
> * Does CN need three different keywords for boolean (`bool`, `boolean` and `CN_bool`). Could we standardize on `bool`, which is what the pretty-printer emits?

Three does seem over the top. I'm not sure what the status of `CN_`-prefixed names is, this has/had some use for how CN fits into the Cerberus pipeline, but maybe this can go away now (similarly `cn_function`).

RE `bool` and `boolean`. `bool` is shorter and nicer, but something that's awkward here is that you can have the C-type `bool` that isn't the same as actual booleans but the integer-valued type. So then if the CN (proper boolean) type is called `bool`, but the same C file talks about C integer type bool, this is quite confusing. 

You do want to have proper booleans in your specification language, so we should probably pick a name in CN that doesn't clash with the name of the C-type. Not clear whether `boolean` solves the problem though.

> * Is the `set` keyword used? Could we deprecate it?

We haven't used it in examples yet, but we should keep `set` as a keyword: unlike lists, sets can't be conveniently defined using inductive datatypes, so they have to be built in.

> * Could we remove the `CN_list` type? What about the `List` constructor for the `BT` type?

We need the `List` constructor in `BT`: the Cerberus C-to-Core elaboration internally introduces lists, so CN needs the `list` type to typecheck those. Given this, I'm actually starting to wonder whether it's not cleaner to keep `list` in the frontend after all: perhaps we should make that machinery work properly and encourage users to use the built-in list type rather than defining their own each time? We also currently don't have machinery for user-defined polymorphic datatypes, such as lists. So the `datatype` a user can write for lists isn't as good as the built-in one in that way. (Sorry, hadn't thought of that.)

@bcpierce00 What's the cost of keeping `list` as a keyword, is this a problem in the naming scheme you're designing? If users were encouraged to use that built-in list type, would it still be a problem?

> * For `tuple` / `pack` / `unpack` - I guess you want to keep the code implementing this functionality in place, and just disable parsing of the keywords?

For `pack` and `unpack` yes. 

For tuple, that's used Cerberus+CN-internally in Core and we need to keep it. Whether we want to keep it in the frontend is a question: the user-accessible language is simpler without; perhaps we want to encourage the user to write records instead of tuples; but sometimes tuples are more convenient than records.

bcpierce00:
>
> You do want to have proper booleans in your specification language, so we
> should probably pick a name in CN that doesn't clash with the name of the
> C-type. Not clear whether boolean solves the problem though.
>
Just to make sure I understand the issues here: Is it 100% certain that the
spec language has to have its own boolean type?  What if we gave the CN
"if" construct the same semantics as in C (zero branch and nonzero
branch)?  Would this be terrible?

> @bcpierce00
<https://urldefense.com/v3/__https://github.com/bcpierce00__;!!IBzWLUs!TOQtwKdX3wWtsxh55hpjiG9CYPWss9CCUebIT1KuTxU2tZsM6R9d3-K5lC-ZbqZlztDoy8gr21s2JjDP5AEdWC-CnvfP$>
What's the cost of keeping list as a keyword, is this a problem in the
naming scheme you're designing? If users were encouraged to use that
built-in list type, would it still be a problem?

The main issue is capitalization.  The current proposed scheme recommends
uc-initial for CN variables, functions, and predicates, and using "list"
would clash very badly with this.  I'd be happier with List.

(I have nearly finished converting the tutorial over to the new scheme --
hope to have something to show later today.)

septract:
> Three does seem over the top. I'm not sure what the status of `CN_`-prefixed names is, this has/had some use for how CN fits into the Cerberus pipeline, but maybe this can go away now (similarly `cn_function`).
> 
> RE `bool` and `boolean`. `bool` is shorter and nicer, but something that's awkward here is that you can have the C-type `bool` that isn't the same as actual booleans but the integer-valued type. So then if the CN (proper boolean) type is called `bool`, but the same C file talks about C integer type bool, this is quite confusing.

I suggest standardizing on `boolean` in that case (assuming per @bcpierce00 we actually need a spec-level boolean type). I propose we deprecate the other two, and update the pretty-printer accordingly. 

(This would require some changes to our testcases I expect). 

> > * Is the `set` keyword used? Could we deprecate it?
> 
> We haven't used it in examples yet, but we should keep `set` as a keyword: unlike lists, sets can't be conveniently defined using inductive datatypes, so they have to be built in.

It seems we don't have any examples of use, is that right? In that case, I suggest we add `set` to our depreciated list until we actually have a use-case. So we would keep the keyword reserved, but discourage users from using a feature that is poorly tested. 

> > * Could we remove the `CN_list` type? What about the `List` constructor for the `BT` type?
> 
> We need the `List` constructor in `BT`: the Cerberus C-to-Core elaboration internally introduces lists, so CN needs the `list` type to typecheck those. Given this, I'm actually starting to wonder whether it's not cleaner to keep `list` in the frontend after all: perhaps we should make that machinery work properly and encourage users to use the built-in list type rather than defining their own each time? We also currently don't have machinery for user-defined polymorphic datatypes, such as lists. So the `datatype` a user can write for lists isn't as good as the built-in one in that way. (Sorry, hadn't thought of that.)

Again, in that case I would suggest we deprecate the keyword until we actually have an example of use. This keeps the keyword available for the future, but discourages use of under-tested features. 

It's a separate question what we want to call the reserved keyword - `list` / `List` per @bcpierce00 / `CN_list` or something else ...

> > * For `tuple` / `pack` / `unpack` - I guess you want to keep the code implementing this functionality in place, and just disable parsing of the keywords?
> 
> For `pack` and `unpack` yes.
> 
> For tuple, that's used Cerberus+CN-internally in Core and we need to keep it. Whether we want to keep it in the frontend is a question: the user-accessible language is simpler without; perhaps we want to encourage the user to write records instead of tuples; but sometimes tuples are more convenient than records.

Again, in that case I suggest deprecating but keeping the keyword `tuple` reserved.

septract:
Also, I missed this comment: 

> similarly `cn_function`

I guess we should remove the `cn_function` keyword too, in that case? I actually have a testcase for `cn_function` [here](https://github.com/rems-project/cn-tutorial/blob/35d58717cea2bc6a6f97df2c96189e850dfe73bb/src/example-archive/simple-examples/working/cn_function_1.c#L12) but if it's mostly unsupported, we should make that clear to users. 

It is super-confusing as a user when a language has a feature but it only semi-works, in my opinion.

cp526:
> You do want to have proper booleans in your specification language, so we should probably pick a name in CN that doesn't clash with the name of the C-type. Not clear whether boolean solves the problem though.
> Just to make sure I understand the issues here: Is it 100% certain that the spec language has to have its own boolean type? What if we gave the CN "if" construct the same semantics as in C (zero branch and nonzero branch)? Would this be terrible?

I think that would be a bit terrible: it would be replicating what I'd think is a design mistake in the C language into CN just to align them more closely, but at the cost of making it harder to understand what specifications actually mean (and what we've verified). Plus all that complexity would then make it into Coq proof of lemmas (whose semantics would have to be in line with CN's own) and any logical reasoning we do above CN-verified statements.

> @bcpierce00
> <https://urldefense.com/v3/__https://github.com/bcpierce00__;!!IBzWLUs!TOQtwKdX3wWtsxh55hpjiG9CYPWss9CCUebIT1KuTxU2tZsM6R9d3-K5lC-ZbqZlztDoy8gr21s2JjDP5AEdWC-CnvfP$> What's the cost of keeping list as a keyword, is this a problem in the naming scheme you're designing? If users were encouraged to use that built-in list type, would it still be a problem? The main issue is capitalization. The current proposed scheme recommends uc-initial for CN variables, functions, and predicates, and using "list" would clash very badly with this. I'd be happier with List. (I have nearly finished converting the tutorial over to the new scheme -- hope to have something to show later today.)

Don't have a very strong opinion about `List` vs `list`, but if C types end up being a subset of the C types (discussed in that separate issue for bounded integer types), does `List` not make CN types look internally inconsistent?

bcpierce00:
>
>
> Again, in that case I would suggest we deprecate the keyword until we
> actually have an example of use. This keeps the keyword available for the
> future, but discourages use of under-tested features.
>
> It's a separate question what we want to call the reserved keyword - list
> / List per @bcpierce00
> <https://urldefense.com/v3/__https://github.com/bcpierce00__;!!IBzWLUs!UJbTCuYg0S71Of0OdCdbtOUUftTReaLneoZK6N1KopQzdWKoeU6fCwZPostwYdGpyGbUAztQi-n559lZw1Jc2FsmaevX$>
> / CN_list or something else ...
>
Deprecating `list` is OK by me, as long as people are happy with the new
naming conventions (I'm using `sllist` for the singly-linked C
implementation of the abstract List type).  But let's not do anything with
`List` for now: Replacing the current tutorial definition with a
polymorphic builtin will be backward compatible.

bcpierce00:
>
> It is super-confusing as a user when a language has a feature but it only
> semi-works, in my opinion.
>
Agreed.  (And this supports my suggestion that the reference grammar should
include pointers to the tutorial, at least until someone writes/updates a
real reference manual.)

cp526:
> Also, I missed this comment:
> 
> > similarly `cn_function`
> 
> I guess we should remove the `cn_function` keyword too, in that case? I actually have a testcase for `cn_function` [here](https://github.com/rems-project/cn-tutorial/blob/35d58717cea2bc6a6f97df2c96189e850dfe73bb/src/example-archive/simple-examples/working/cn_function_1.c#L12) but if it's mostly unsupported, we should make that clear to users.
> 
> It is super-confusing as a user when a language has a feature but it only semi-works, in my opinion.



> Also, I missed this comment:
> 
> > similarly `cn_function`
> 
> I guess we should remove the `cn_function` keyword too, in that case? I actually have a testcase for `cn_function` [here](https://github.com/rems-project/cn-tutorial/blob/35d58717cea2bc6a6f97df2c96189e850dfe73bb/src/example-archive/simple-examples/working/cn_function_1.c#L12) but if it's mostly unsupported, we should make that clear to users.
> 
> It is super-confusing as a user when a language has a feature but it only semi-works, in my opinion.

Sorry, I was confused. This is actually a separate feature from what I thought. It's a way to automatically derive a CN function from a pure C function, which currently only works for very simple definitions, so shouldn't be "advertised" in the tutorial or language reference, but is used in some pKVM verification examples.

bcpierce00:
>
> Sorry, I was confused. This is actually a separate feature from what I
> thought. It's a way to automatically derive a CN function from a pure C
> function, which currently only works for very simple definitions, so
> shouldn't be "advertised" in the tutorial or language reference, but is
> used in some pKVM verification examples.
>
> The reference grammar should mark such features as "experimental / under
construction / whatever"

cp526:
Both `list` and `tuple` are tested, because they're used internally, so there's no reason to deprecate those.

And for sets, it's not unlikely those will soon come up in the pKVM pagetable verification, so I'd rather not deprecate those either. We could just make `compile.ml` or `wellTyped.ml` warn the user that those are experimental if we want to discourage outside users from trying them.

septract:
Okay perhaps there are three categories of keywords - "production" / "experimental" (generates a warning) / "deprecated" (generates an error). 

When you say tuple / list are used 'internally', do you mean "internally in the CN team's examples", or "internally in the tool but not in any example code"

cp526:
> Okay perhaps there are three categories of keywords - "production" / "experimental" (generates a warning) / "deprecated" (generates an error).

Agreed. Types `list`, `tuple`, and `set` could then all go into the `experimental` category (modulo any renaming for consistency with the naming scheme).

> 
> When you say tuple / list are used 'internally', do you mean "internally in the CN team's examples", or "internally in the tool but not in any example code"

The latter, so the internal typing logic is tested, but not the frontend.

yav:
One point about keeping `list` is that it is polymorphic, and I don't think we support polymorphic datatype declarations at the moment or am I confused?

septract:
Great, so I think we have converged on a plan for handling keywords. I will add an extra category of keywords, `experimental`. The semantics of the categories is as follows: 
   * `production` - works as normal 
   * `experimental` - generate a warning (`list`, `tuple`, `cn_function`, `set` - I will see if there are others) 
   * `unimplemented` - a reserved keyword, but generates an error and refuse to compile (`pack`, `unpack`). 

I'm not sure we have converged on renaming `list`. @bcpierce00 it sounds like you would prefer we change the keyword to `List`, is this correct? Does that apply to `set`, `tuple` etc?

bcpierce00:
>
> I'm not sure we have converged on renaming list. @bcpierce00
> <https://urldefense.com/v3/__https://github.com/bcpierce00__;!!IBzWLUs!VHF5IrUbbKikRvktuqiLiu2pujhPYDWxOfB2rrdH-Joz4T3Dl34NEPLNzbNgSncNuMOrZhAKF928w8CPY5_9ju7CiEXq$>
> it sounds like you would prefer we change the keyword to List, is this
> correct? Does that apply to set, tuple etc?
>
Yes, probably.  But

1) This conversation will make more sense once we've got a naming
convention that people are reasonably happy with.
2) I don't want List to become reserved until I can throw away my personal
definition and just use the built-in one.

cp526:
@septract Sounds good.

@bcpierce00 We could do `cn_list`, `cn_tuple`, `cn_set` until we settle on the exact names and those don't clash with existing examples any longer? Given we're marking those types experimental the awkward prefixed names are good enough for now, I'd say.

septract:
This is still a bit broken, but ready for an initial review @cp526

Issues:
* Some mystery parsing problems on `Owned<bool>` 
* Need to fix the `rems-project/cn-tutorial` test-cases to match the new syntax, mostly for `bool` vs. `boolean`

ocaml_frontend/pprinters/pp_errors.ml
```diff
@@ -56,6 +56,10 @@ let string_of_cparser_cause = function
       "unterminated " ^ str ^ " comment"
   | Cparser_mismatched_magic_comment ->
       "magic comment syntax at open but not close"
+  | Cparser_experimental_keyword str -> 
+      "experimental keyword '" ^ str ^ "'. Use of experimental features is discouraged"
```
galoiscole:
```suggestion
      "experimental keyword '" ^ str ^ "' (use of experimental features is discouraged)"
```

parsers/c/c_lexer.mll
```diff
@@ -568,9 +605,7 @@ and initial flags = parse
         Hashtbl.find lexicon id
       with Not_found ->
         if flags.inside_cn then
-          try Hashtbl.find cn_lexicon id
-          with Not_found ->
-            LNAME id
```
cp526, line 573: The LNAME is getting lost: cn_lex_keyword is always returning a UNAME in the ultimate Not_found case, so it probably has to parameterised what it returns then. Perhaps related to the `bool` problem?

(UNAME: uppercase name, LNAME: lowercase name)

parsers/c/c_lexer.mll
```diff
@@ -152,10 +145,56 @@ let cn_keywords: (string * Tokens.token) list = [
...
+         Errors.(CPARSER (Errors.Cparser_experimental_keyword id))
+         Warning);
+      kw 
+    with Not_found ->
+      try 
+        (* Try to lex CN unimplemented keywords - throw an error on success *)
+        let _ = Hashtbl.find (cn_lex_builder cn_keywords_unimplemented) id in 
+        raise (Error (Errors.Cparser_unimplemented_keyword id))
+      with Not_found ->
+        UNAME id
```
cp526: I wonder whether the three-level lookup will be slow. If it is, we may want to have a single keywords list of type `(string * (Tokens.token * (normal+experimental+unimplemented) ) list`

cp526:
Looks good to me, except for minor comments (inline) and one other point: `map` shouldn't be marked experimental (and not renamed `cn_map`), we've used that extensively in the buddy allocator verification, so it's well tested. It also shows up as soon as you're working with arrays.

(The lexer Kayvan or Dhruv may know better.)

parsers/c/c_lexer.mll
septract: Fixed

parsers/c/c_lexer.mll
septract: I haven't benchmarked this, but this seems like a small source of inefficiency so I'm inclined to let it go. Would you prefer if I refactored it?

septract: Actually, I think your suggestion is better - I've implemented it.

septract:
I addressed all the comments above, and indeed, it looks like the source of the parsing issues was the bug @cp526 pointed out. 

Ready to merge, unless there are other comments. Note that we need to merge https://github.com/rems-project/cn-tutorial/pull/76 before this PR to avoid a CI failure.

parsers/c/c_lexer.mll
```diff
@@ -92,70 +92,98 @@ let lexicon: (string, token) Hashtbl.t =
...
+  * 'experimental' - functional in some cases but not recommended for general use 
+  * 'unimplemented' - non-functional, but the keyword is reserved 
+
+May raise `Not_found`, indicating `id` is not a recognized CN keyword. *)
+let cn_lex_keyword id start_pos end_pos = 
+  (* Try to lex CN production keywords *)
+  let cn_lex = Hashtbl.create 0 in
+  let add (key, builder) = Hashtbl.add cn_lex key builder in
+  List.iter add cn_keywords; 
+  match Hashtbl.find cn_lex id with
```
dc-mak, line 176: @cp526 @septract this is recreating the Hashtable on each call to look up a keyword; probably want something like the below...
```ocaml
let cn_lex_keyword = 
  (* Try to lex CN production keywords *)
  let cn_lex = Hashtbl.create 0 in
  let add (key, builder) = Hashtbl.add cn_lex key builder in
  List.iter add cn_keywords;
  fun id start_pos end_pos ->
  match Hashtbl.find cn_lex id with 
  ```

PR #504, opened 2024-08-16
Prove `amap_add_list_not_at` with fixed formulation

zoickx:
Original formulation of the lemma proven incorrect in 95e50c3fbd8b15b3f74dff638ba90eb4a33274de.
Fixed to guard against overflow, all dependent proofs updated accordingly.


zoickx:
@vzaliva

PR #518, opened 2024-08-20
Update cn-runtime-dir script

rbanerjee20:
Updated `tests/cn-runtime/cn-runtime-dir.sh` to match [`runtime-test.sh`](https://github.com/rems-project/cn-tutorial/blob/cn-runtime-testing/runtime-test.sh) from the CN tutorial repo, which is also what is being used in the runtime testing CI right now, since this script is cleaner to work with and shows test success/failure more clearly in the terminal.

### DIFF
--- a/tests/cn-runtime/cn-runtime-dir.sh
+++ b/tests/cn-runtime/cn-runtime-dir.sh
@@ -41,7 +41,6 @@ SUCCESS=$(find tests/cn -name '*.c' \
 BUGGY=("")
 
 SHOULD_FAIL=$(find tests/cn -name '*.error.c')
-# SHOULD_FAIL=("src/examples/read.broken.c" "src/examples/slf14_basic_succ_using_incr_attempt.broken.c")
 
 FAILED=""
 
--- a/tests/cn-runtime/cn-runtime-dir.sh
+++ b/tests/cn-runtime/cn-runtime-dir.sh
@@ -1,97 +1,72 @@
-#! /bin/bash
+#!/usr/bin/env bash
+set -euo pipefail -o noclobber
 
-# tests for the CN executable spec tool attached to cerberus
+function echo_and_err() {
+    printf "$1\n"
+    exit 1
+}
 
-DIRNAME=$1
+[ $# -eq 0 ] || echo_and_err "USAGE: $0"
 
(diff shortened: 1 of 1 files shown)


tests/cn-runtime/cn-runtime-dir.sh
```diff
@@ -1,97 +1,71 @@
...
+RUNTIME_PREFIX="$OPAM_SWITCH_PREFIX/lib/cn/runtime"
+[ -d "${RUNTIME_PREFIX}" ] || echo_and_err "Could not find CN's runtime directory (looked at: '${RUNTIME_PREFIX}')"
 
-NUM_GENERATION_FAILED=0
-GENERATION_FAILED=''
+CHECK_SCRIPT="${RUNTIME_PREFIX}/libexec/cn-runtime-single-file.sh"
 
-NUM_COMPILATION_FAILED=0
-COMPILATION_FAILED=''
+[ -f "${CHECK_SCRIPT}" ] || echo_and_err "Could not find single file helper script: ${CHECK_SCRIPT}"
```
dc-mak, line 16: I think this is a copy paste error.

rbanerjee20: I copied the script in its entirety and this line looks the same in the original ([here](https://github.com/rems-project/cn-tutorial/blob/c6a34f25e8db6a6198c13846d4f79c8ed1f546c9/runtime-test.sh#L16)) - unless you think this is a problem in the original as well?

rbanerjee20: I also tried running the script locally and it seems to be working well

dc-mak: No as in, this is the "dir" script but the string says "single file". My bash is impeccable of course ;-)

rbanerjee20: Ah, got it. No, this is correct because it's looking for the single file helper script within the Cerberus runtime library regardless of what the top-level script (in this case, the `cn-runtime-dir` script I pushed) is doing. It's `cn-runtime-single-file` which is being searched for in `$OPAM_SWITCH_PREFIX/lib/cn/runtime/libexec` and if it's not found, this is the error that gets thrown.

PR #522, opened 2024-08-20
[CN] Fix #380 (Have `ptr_eq` typechecks its arguments)

### LINKED ISSUE #380: [CN] `ptr_eq` typing (confusingly) not strict enough
Specifically, `x` and `y` should have type `u64` not `pointer`

```
12:56 ➜  2024-cn-testing-paper git:(main) cat tmp.c && cn tmp.c
void f(unsigned int x, unsigned int y)
/*@ requires ptr_eq(x,y);
   ensures true; @*/
{ }

[1/1]: f
```

dsainati1:
Fixes https://github.com/rems-project/cerberus/issues/380

### DIFF
--- a/backend/cn/lib/builtins.ml
+++ b/backend/cn/lib/builtins.ml
@@ -5,14 +5,13 @@ open Effectful.Make (Resultat)
 
 open TypeErrors
 open IndexTerms
+module CF = Cerb_frontend
+
+type builtin_fn_def = string * Sym.t * LogicalFunctions.definition
+
+let loc = Cerb_location.other "<builtin>"
 
 (* builtin function symbols *)
-let mk_arg0 mk args loc =
-  match args with
-  | [] -> return (mk loc)
-  | _ :: _ as xs ->
-    fail { loc; msg = Number_arguments { has = List.length xs; expect = 0 } }
-
 
 let mk_arg1 mk args loc =
   match args with
@@ -36,10 +35,28 @@ let mk_arg3_err mk args loc =
 
 let mk_arg3 mk = mk_arg3_err (fun tup loc -> return (mk tup loc))
 
-let mk_arg5 mk args loc =
-  match args with
-  | [ a; b; c; d; e ] -> return (mk (a, b, c, d, e) loc)
-  | xs -> fail { loc; msg = Number_arguments { has = List.length xs; expect = 5 } }
+let var_binop op ty ~left:(sym1, bt1) ~right:(sym2, bt2) =
+  binop op (sym_ (sym1, bt1, loc), sym_ (sym2, bt2, loc)) loc ty
+
+
+let definition name args body =
+  ( name,
+    Sym.fresh_named name,
+    LogicalFunctions.
+      { loc; emit_coq = false; args; definition = Def body; return_bt = bt body } )
+
+
+let mk_builtin_arg0 name = definition name []
+
+let mk_builtin_arg1 name bt mk : builtin_fn_def =
+  let arg = (Sym.fresh_named "arg", bt) in
+  mk arg |> definition name [ arg ]
+
+
+let mk_builtin_arg2 name (bt1, bt2) mk : builtin_fn_def =
+  let left = (Sym.fresh_named "arg1", bt1) in
+  let right = (Sym.fresh_named "arg2", bt2) in
+  mk ~left ~right |> definition name [ left; right ]
 
 
 let min_bits_def (sign, n) =
@@ -49,9 +66,7 @@ let min_bits_def (sign, n) =
     | Signed -> (Z.(neg @@ shift_left one (Int.sub n 1)), "i")
   in
   let name = "MIN" ^ letter ^ Int.to_string n in
-  ( name,
-    Sym.fresh_named name,
-    mk_arg0 (fun loc -> IT.sterm_of_term @@ num_lit_ num (BT.Bits (sign, n)) loc) )
+  num_lit_ num (BT.Bits (sign, n)) loc |> mk_builtin_arg0 name
 
 
 let max_bits_def (sign, n) =
@@ -61,11 +76,65 @@ let max_bits_def (sign, n) =
     | Signed -> (Z.(shift_left one (Int.sub n 1) - one), "i")
   in
   let name = "MAX" ^ letter ^ Int.to_string n in
-  ( name,
-    Sym.fresh_named name,
-    mk_arg0 (fun loc -> IT.sterm_of_term @@ num_lit_ num (BT.Bits (sign, n)) loc) )
+  num_lit_ num (BT.Bits (sign, n)) loc |> mk_builtin_arg0 name
 
 
+let max_min_bits =
+  let sizes = [ 8; 16; 32; 64 ] in
+  let signs = [ BT.Unsigned; Signed ] in
+  List.fold_left
+    (fun acc sign ->
+      List.fold_left
+        (fun acc size -> max_bits_def (sign, size) :: min_bits_def (sign, size) :: acc)
+        acc
+        sizes)
+    []
+    signs
+
+
+let not_def =
+  mk_builtin_arg1 "not" BT.Bool (fun (sym, bt) -> not_ (sym_ (sym, bt, loc)) loc)
+
+
+let is_null_def : builtin_fn_def =
+  mk_builtin_arg1 "is_null" BT.Loc (fun (sym, bt) ->
+    (eq__ (sym_ (sym, bt, loc)) (null_ loc)) loc)
+
+
+(* Cannot translate this to a logical function until the TODO in `cn_to_ail_expr_aux_internal` in `cn_internal_to_ail.ml` is resolved*)
+let has_alloc_id_def =
+  ( "has_alloc_id",
+    Sym.fresh_named "has_alloc_id",
+    mk_arg1 (fun p loc' -> IT.sterm_of_term @@ IT.hasAllocId_ (IT.term_of_sterm p) loc')
+  )
+
+
+let ptr_eq_def : builtin_fn_def =
+  var_binop EQ BT.Bool |> mk_builtin_arg2 "ptr_eq" (BT.Loc, BT.Loc)
+
+
+let prov_eq_def : builtin_fn_def =
+  let left = (Sym.fresh_named "arg1", BT.Loc) in
+  let right = (Sym.fresh_named "arg2", BT.Loc) in
+  let left_cast = allocId_ (sym_ (fst left, BT.Loc, loc)) loc in
+  let right_cast = allocId_ (sym_ (fst right, BT.Loc, loc)) loc in
+  let body = binop EQ (left_cast, right_cast) loc BT.Bool in
+  definition "prov_eq" [ left; right ] body
+
+
+let addr_eq_def : builtin_fn_def =
+  let left = (Sym.fresh_named "arg1", BT.Loc) in
+  let right = (Sym.fresh_named "arg2", BT.Loc) in
+  let left_cast = addr_ (sym_ (fst left, BT.Loc, loc)) loc in
+  let right_cast = addr_ (sym_ (fst right, BT.Loc, loc)) loc in
+  let body = binop EQ (left_cast, right_cast) loc BT.Bool in
+  definition "addr_eq" [ left; right ] body
+
+
+(* The remaining functions in this file, from here until array_to_list_def cannot yet be translated to
+   LogicalFunction.definition types because they implicitly require basetype polymorphism.
+   For example, the `mod` function allows inputs of any sign and size, but such a function cannot be defined
+   yet with an index term *)
 let mul_uf_def = ("mul_uf", Sym.fresh_named "mul_uf", mk_arg2 mul_no_smt_)
 
 let div_uf_def = ("div_uf", Sym.fresh_named "div_uf", mk_arg2 div_no_smt_)
@@ -114,8 +183,6 @@ let rem_def = ("rem", Sym.fresh_named "rem", mk_arg2 rem_)
 
 let mod_def = ("mod", Sym.fresh_named "mod", mk_arg2 mod_)
 
-let not_def = ("not", Sym.fresh_named "not", mk_arg1 not_)
-
 let nth_list_def = ("nth_list", Sym.fresh_named "nth_list", mk_arg3 nthList_)
 
 let array_to_list_def =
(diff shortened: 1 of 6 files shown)


backend/cn/lib/core_to_mucore.ml
```diff
@@ -1527,8 +1527,16 @@ let normalise_file ~inherit_loc ((fin_markers_env : CAE.fin_markers_env), ail_pr
...
   let@ env = C.add_datatype_infos env ail_prog.cn_datatypes in
-  let@ env = C.register_cn_functions env ail_prog.cn_functions in
-  let@ lfuns = ListM.mapM (C.translate_cn_function env) ail_prog.cn_functions in
+  let cn_funs = Builtins.builtin_cn_funs @ ail_prog.cn_functions in
+  let@ env = C.register_cn_functions env cn_funs in
+  let@ lfuns =
+    ListM.mapiM
+      (fun index ->
+        C.translate_cn_function
+          ~builtin:(index < List.length Builtins.builtin_cn_funs)
```
dc-mak: `List.length` is a linear time function called on each iteration here. Perhaps simpler to just have two calls to translate_cn_function, one for the builtins, and another for ail_prog.cn_functions?

dsainati1: Right yeah I forgot `length` works like this in OCaml

backend/cn/lib/builtins.ml
```diff
@@ -135,41 +167,66 @@ let array_to_list_def =
...
+        (CNExpr
+           ( builtin_location,
+             CNExpr_binop
+               ( CN_equal,
+                 CNExpr (builtin_location, CNExpr_var p1),
+                 CNExpr (builtin_location, CNExpr_var p2) ) )))
+    CN_bool
 
 
 let prov_eq_def =
```
dsainati1: This `prov_eq` function wasn't used anywhere in the tests. What is the difference between this, `ptr_eq` and `addr_eq`

dc-mak: It's related to memory-model aspects of the CN (in case you're unfamiliar: a memory model is the part of the contract between programmers and compiler writers which sets out the rules for the latter to follow, and defines the envelope of optimisations for the former to engage in, relating to the rules around creating, manipulating and destroying pointers and memory). A memory model of C that seems to have some consensus is called PNVI-ae-udi, but the one CN will use (as soon as I'm done implementing it) is based on VIP.

If you can keep it for now, that would be helpful, because I may need it once I finish up other implementations and turn on VIP by default. You can put a comment next to it saying it should be deleted if it's not used after CN switches to enabling VIP by default.

backend/cn/lib/builtins.ml
```diff
@@ -36,12 +41,52 @@ let mk_arg3_err mk args loc =
...
+        cn_func_attrs = [];
+        cn_func_args = [ (arg1, t1); (arg2, t2) ];
+        cn_func_body = mk (arg1, arg2);
+        cn_func_return_bty = return
+      } )
 
 
+let arith_binop op (it, it') loc = IT (Binop (op, it, it'), bt it, loc)
+
 let min_bits_def (sign, n) =
```
dsainati1, line 62: A lot of these functions remain to be translated into the new `builtin_fn` type, but not all of them seem possible to do at this time, for a couple reasons:

1) Some of the binops supported in the `IT.binop` type don't have equivalents in the `CN.binop` type (e.g. `ShiftLeft` and `ShiftRight`), so it is unclear to me how to write definitions for these.

2) Some of the binops that are supported (e.g. `BW_And` and `BW_Or`) appear to be functionally polymorphic, in that they operate on unsigned integers of multiple different sizes. Is it possible to specify a function definition that is polymorphic in this way in the CN intermediate language?

dc-mak: Aaargh, that is a very good point I should have thought about earlier. I think this means that for now, the way to proceed would be to go back to the original plan. Sorry for the backtracking!

dsainati1: I reverted back to the original plan which addressed issue 1 here, but issue 2 here remains an issue: some of these builtins operate over all types of `int`s and `uint`s, rather than one specific input type. For something like `ptr_eq` we can just create a function with two `pointer` arguments whose body is an index term comparing their value, but for `mod`, for example, it does not seem possible to define a concrete `LogicalFunction.definition` that permits multiple types of arguments (e.g. a `u32` and a `u64` both).

dc-mak: In that case, can you convert the ones you can and leave a comment that we need basetype polymorphism for the ones leftover. It's a bit ugly and unfortunate, but something is better than nothing here.

backend/cn/lib/builtins.ml
```diff
@@ -5,14 +5,13 @@ open Effectful.Make (Resultat)
...
 
-(* builtin function symbols *)
-let mk_arg0 mk args loc =
-  match args with
-  | [] -> return (mk loc)
-  | _ :: _ as xs ->
-    fail { loc; msg = Number_arguments { has = List.length xs; expect = 0 } }
+type builtin_fn = LogicalFunctions.definition
 
+let builtin_location = Cerb_location.other "<builtin>"
```
dc-mak: This can just be called `loc` for short - the context will make it clear, and it will allow for record punning in records.

backend/cn/lib/builtins.ml
```diff
@@ -36,10 +35,37 @@ let mk_arg3_err mk args loc =
...
-let mk_arg5 mk args loc =
-  match args with
-  | [ a; b; c; d; e ] -> return (mk (a, b, c, d, e) loc)
-  | xs -> fail { loc; msg = Number_arguments { has = List.length xs; expect = 5 } }
+let var_binop op ty (lhs_sym, lhs_t) (rhs_sym, rhs_t) =
+  binop
+    op
+    (sym_ (lhs_sym, lhs_t, builtin_location), sym_ (rhs_sym, rhs_t, builtin_location))
+    builtin_location
+    ty
```
dc-mak: We usually use `bt` or `sbt` to name types; you can also use labelled arguments to distinguish left and right, so for example:
```ocaml
let var_binop op ret_bt ~left:(sym1, bt1) ~right:(sym2, bt2) =
```

backend/cn/lib/builtins.ml
```diff
@@ -61,11 +85,48 @@ let max_bits_def (sign, n) =
...
+  mk_builtin_arg1 "is_null" BT.Loc (fun (sym, t) ->
+    (sym_ (sym, t, builtin_location) %== null_ builtin_location) builtin_location)
+
+
+let ptr_eq_def : string * Sym.t * builtin_fn =
+  var_binop EQ BT.Bool |> mk_builtin_arg2 "ptr_eq" (BT.Loc, BT.Loc)
+
+
+let prov_eq_def : string * Sym.t * builtin_fn =
+  var_binop EQ BT.Bool |> mk_builtin_arg2 "prov_eq" (BT.Loc, BT.Loc)
```
dc-mak: The types are correct but the implementation needs to cast the arguments to `allocId_` before calling equal.

backend/cn/lib/builtins.ml
```diff
@@ -61,11 +85,48 @@ let max_bits_def (sign, n) =
...
+
+
+let addr_eq_def : string * Sym.t * builtin_fn =
+  var_binop EQ BT.Bool |> mk_builtin_arg2 "addr_eq" (BT.Loc, BT.Loc)
```
dc-mak: Similarly, the types are correct but the implementation needs to cast the arguments with `addr_` before calling equal.

backend/cn/lib/builtins.ml
```diff
@@ -222,4 +232,8 @@ let apply_builtin_funs fsym args loc =
     return (Some t)
 
 
-let cn_builtin_fun_names = List.map (fun (str, sym, _) -> (str, sym)) builtin_funs
+(* This list of names is later passed to the frontend so that these are available in the elaboration,
+   so it should include all builtin function names *)
```
dc-mak: In `bin/main.ml`

backend/cn/lib/compile.ml
```diff
@@ -455,7 +436,7 @@ module EffectfulTranslation = struct
 
   (* TODO: type checks and disambiguation at this stage seems ill-advised,
      ideally would be integrated into wellTyped.ml *)
-  let mk_translate_binop loc' bop (e1, e2) =
+  let mk_translate_binop ?(builtin = false) loc' bop (e1, e2) =
```
dc-mak: I believe `builtin` should be unnecessary given the change in approach.

backend/cn/lib/compile.ml
```diff
@@ -24,11 +24,6 @@ module STermMap = Map.Make (struct
 module StringMap = Map.Make (String)
 module StringSet = Set.Make (String)
 
-type function_sig =
-  { args : (Sym.t * BaseTypes.t) list;
-    return_bty : BaseTypes.t
-  }
-
```
dc-mak: It would be cleaner (I think) to just keep this type and convert the Builtins to this rather than converting giving dummy definitions to other functions, though if this is not the case I would like to know.

backend/cn/lib/builtins.ml
```diff
@@ -61,11 +85,48 @@ let max_bits_def (sign, n) =
...
+
 
+let not_def =
+  mk_builtin_arg1 "not" BT.Bool (fun (sym, t) ->
+    not_ (sym_ (sym, t, builtin_location)) builtin_location)
 
+
+let is_null_def : string * Sym.t * builtin_fn =
+  mk_builtin_arg1 "is_null" BT.Loc (fun (sym, t) ->
+    (sym_ (sym, t, builtin_location) %== null_ builtin_location) builtin_location)
```
dc-mak: We don't really use infix operators in the code base yet, so I would recommend deleting this (and its definition).

dc-mak (changes requested):
On the right track overall, thank you. Should be ready to merge in a couple of iterations I think.

backend/cn/lib/builtins.ml
dsainati1: I found an `Alloc_id` base type, but which base type does `addr_` refer to?

backend/cn/lib/compile.ml
dsainati1: I could be misunderstanding something, but if we do this, where are the actual implementations of the builtins used?

dc-mak: In `core_to_mucore.ml`, logical functions from the frontend are converted to `lfuns` via `translate_cn_function` and then put in the terribly named and should be renamed `mu_logical_predicates`. Those are then read in `check.ml` by being passed to `record_and_check_logical_functions`. From there, they are added (first without definitions, for mutual recursion), type-checked, and then have there definitions added in after type-checking. If the definition is needed, then `get_logical_function_def` is called on the `Global.t` typing environment.

`register_cn_function` should only add the signatures (so this can handle builtin and user-defined functions at the same time) but `translate_cn_function` converts the definitions for user-defined functions, you just prepend the builtins to `lfuns` and the rest will work out fine.

backend/cn/lib/builtins.ml
dc-mak: `Memory.uintptr_bt`

backend/cn/lib/builtins.ml
```diff
@@ -120,7 +120,12 @@ let prov_eq_def : builtin_fn_def =
 
 
 let addr_eq_def : builtin_fn_def =
-  var_binop EQ BT.Bool |> mk_builtin_arg2 "addr_eq" (BT.Loc, BT.Loc)
+  let left = (Sym.fresh_named "arg1", BT.Loc) in
+  let right = (Sym.fresh_named "arg2", BT.Loc) in
+  let left_cast = cast_ Memory.uintptr_bt (sym_ (fst left, BT.Loc, loc)) loc in
+  let right_cast = cast_ Memory.uintptr_bt (sym_ (fst right, BT.Loc, loc)) loc in
```
dc-mak: There's a `IT.addr_` which does the cast for you.

backend/cn/lib/builtins.ml
```diff
@@ -61,11 +76,62 @@ let max_bits_def (sign, n) =
...
+let ptr_eq_def : builtin_fn_def =
+  var_binop EQ BT.Bool |> mk_builtin_arg2 "ptr_eq" (BT.Loc, BT.Loc)
+
+
+let prov_eq_def : builtin_fn_def =
+  let left = (Sym.fresh_named "arg1", BT.Loc) in
+  let right = (Sym.fresh_named "arg2", BT.Loc) in
+  let left_cast = cast_ BT.Alloc_id (sym_ (fst left, BT.Loc, loc)) loc in
+  let right_cast = cast_ BT.Alloc_id (sym_ (fst right, BT.Loc, loc)) loc in
+  let body = binop EQ (left_cast, right_cast) loc BT.Bool in
```
dc-mak: There's a `IT.allocId_` which does the cast for you.

backend/cn/lib/indexTerms.ml
```diff
@@ -647,8 +649,6 @@ let ( %* ) t t' = mul_ (t, t')
 
 let ( %/ ) t t' = div_ (t, t')
 
-let ( %== ) t t' = eq_ (t, t')
-
```
dc-mak: Ugh I didn't realise there were other operators. If you can put this back for now to make the diff smaller, I can delete all the operators in one go in a separate PR/issue.

backend/cn/lib/core_to_mucore.ml
dc-mak, line 1: Can you add a above the below saying that (a) these are the user-defined functions and (b) some of the builtins are registered in `Compile.init_env`?
```ocaml
  let@ env = C.register_cn_functions env ail_prog.cn_functions in
  ```

**Review by dc-mak — CHANGES_REQUESTED** [review]
Last round of changes - just a few tweaks, thank you so much. If you can rebase and squash that would also be helpful.

**Comment by dc-mak** [comment]
Also once this PR is done and merged, can you do a quick PR adding `.mli` files for `Core_to_mucore` and `Compile`? https://github.com/rems-project/cerberus/blob/master/backend/cn/ONBOARDING.md#adding-mli-files

**Comment by dc-mak** [comment]
There look to be CI failures. Can you please investigate? One of them is code-formatting.
For runtime testing related things, you can ask @rbanerjee20 if you get stuck.

**Review by dc-mak — COMMENTED** [review]
Lgtm but CI is failing.

**Comment by dsainati1** [comment]
Hmm, I don't know exactly what the failing runtime tests are or how to invoke them locally, and the CI job needs to be approved by a maintainer.

**Comment by dc-mak** [comment]
Are you able to view which CI jobs pass or fail? And the logs of those or not?

Regardless, if you look in the workflow file, you'll find the command: https://github.com/rems-project/cerberus/actions/runs/10599783302/workflow?pr=522#L85 : `cd tests && ./run-cn-exec.sh`

Before running the file, you can open it to see `timeout 20 "${CHECK_SCRIPT}" "${SCRIPT_OPT}" "$file" &> /dev/null` called in a function which is called in a loop. Alternatively, you can uncomment `set -xv` for debugging and run the file itself.

It will produce a bunch of output, including something like `/home/dhruv/.opam/4.14.1/lib/cn/runtime/libexec/cn-runtime-single-file.sh -oq tests/cn/missing_resource_indirect.error.c` which can be simplified down to 
`${OPAM_SWITCH_PREFIX}/lib/cn/runtime/libexec/cn-runtime-single-file.sh -oq tests/cn/missing_resource_indirect.error.c` 
or 
`./runtime/libcn/libexec/cn-runtime-single-file.sh -oq tests/cn/missing_resource_indirect.error.c`

We should add this to the ONBOARDING doc.

**Comment by dc-mak** [comment]
If you follow a similar process with the CI build workflow, you'll see that `slf_incr2.c` in the tutorial repo is now failing (because `ptr_eq` is now stricter).

**Comment by dsainati1** [comment]
I submitted https://github.com/rems-project/cn-tutorial/pull/83 to the tutorial repo to fix the issue with `slf_incr2.c`, and fixed the runtime issue I believe.

**Comment by dc-mak** [comment]
The tutorial is still failing for some reason even though your change was merged in https://github.com/rems-project/cn-tutorial/pull/83

The runtime is now showing a further bug - it needs to be told where to find and substitute definitions of the builtins.

**Review by dc-mak — CHANGES_REQUESTED** [review]
See above.

**Comment by rbanerjee20** [comment]
> The runtime is now showing a further bug - it needs to be told where to find and substitute definitions of the builtins.

I'll take a look

**Comment by rbanerjee20** [comment]
This is what I'm getting when running the single-file runtime testing script on `slf_incr2.c`:

```python
slf_incr2.c:19:21: error: Type error
            ptr_eq (vs_.pv,
                    ~~~^~~ 
Expression 'vs_.pv' has type 'u32'.
I expected it to have type 'pointer' because of other location (<builtin>) 

Failed to generate C files from CN-annotatated source.
```

Is this the same error you were talking about @dc-mak ?

dsainati1:
That should be fixed, the tutorial's version of slf_incr2.c should be updated to use `==` instead of `ptr_eq` there.

rbanerjee20:
Ah - the CN runtime testing CI relies on the `cn-runtime-testing` branch of the CN tutorial, not `main`. It will use `main` soon but this is still a WIP. I can update that branch and see if it helps with the Cerberus CI

rbanerjee20:
Great, so runtime testing CI seems to be working again with that same change from the `main` branch of CN tutorial now also applied to `cn-runtime-testing`. I did try the script locally on `slf_incr2.c` and got the following 2 warnings though:

```c
cn.c:43:34: warning: integer literal is too large to be represented in a signed integer type, interpreting as unsigned [-Wimplicitly-unsigned-literal]
  return convert_to_cn_bits_i64(-9223372036854775808);
                                 ^
cn.c:78:33: warning: integer literal is too large to be represented in a signed integer type, interpreting as unsigned [-Wimplicitly-unsigned-literal]
  return convert_to_cn_bits_u64(18446744073709551615);
                                ^
2 warnings generated.

```

Looks a bit suspicious that these kinds of integer literals are being generated - probably worth investigating even though the runtime testing CI is happy now.

rbanerjee20:
And the remaining CI failure seems to be CN test-gen-related - @ZippeyKeys12 would be the right person to ask about this.

dc-mak:
> Great, so runtime testing CI seems to be working again with that same change from the `main` branch of CN tutorial now also applied to `cn-runtime-testing`. I did try the script locally on `slf_incr2.c` and got the following 2 warnings though:
> 
> ```c
> cn.c:43:34: warning: integer literal is too large to be represented in a signed integer type, interpreting as unsigned [-Wimplicitly-unsigned-literal]
>   return convert_to_cn_bits_i64(-9223372036854775808);
>                                  ^
> cn.c:78:33: warning: integer literal is too large to be represented in a signed integer type, interpreting as unsigned [-Wimplicitly-unsigned-literal]
>   return convert_to_cn_bits_u64(18446744073709551615);
>                                 ^
> 2 warnings generated.
> ```
> 
> Looks a bit suspicious that these kinds of integer literals are being generated - probably worth investigating even though the runtime testing CI is happy now.

Known issue: https://github.com/rems-project/cerberus/issues/536

dc-mak:
Superseded by https://github.com/rems-project/cn/pull/4

PR #530, opened 2024-08-23
Add benchmarking CI that generates graphs on updates to master

jprider63:
This PR adds a CI workflow that generates performance graphs on every update to master. It stores the data as json in the `gh-pages` branch and renders the graphs on a Github Page. Here's [an example](https://galoisinc.github.io/cerberus/dev/bench/) of what this looks like with [dummy data](https://github.com/GaloisInc/cerberus/blob/gh-pages/dev/bench/data.js). 

This currently runs `cn` on all the (successful) .c files in the test suite. Eventually, we probably want to make a proper benchmark suite using something like [Core_bench](https://blog.janestreet.com/core_bench-micro-benchmarking-for-ocaml/).

**Before this is merged**, someone with permission needs to create a branch called `gh-pages` with empty contents. You can do this by grabbing the branch [`gh-pages-init`](https://github.com/GaloisInc/cerberus/tree/gh-pages-init) from our fork and pushing to `gh-pages` in this repository. 

Implements part of rems-project/cn-tutorial/issues/54.

### DIFF
--- a/tests/run-ci-benchmarks.sh
+++ b/tests/run-ci-benchmarks.sh
@@ -1,0 +1,55 @@
+#!/bin/bash
+
+JSON_FILE="benchmark-data.json"
+
+cat << EOF >> ${JSON_FILE}
+[
+EOF
+
+DIRNAME=$(dirname "$0")
+
+SUCC=$(find "${DIRNAME}"/cn -name '*.c' | grep -v '\.error\.c')
+
+COUNT=0
+for TEST in ${SUCC}; do
+  let COUNT=${COUNT}+1
+done
+
+INDEX=0
+echo $SUCC
+echo $COUNT
+for TEST in ${SUCC}; do
+  echo ${INDEX}
+
+  # Record wall clock time in seconds
+  /usr/bin/time -f "%e" -o /tmp/time cn verify ${TEST}
+  TIME=$(cat /tmp/time)
+
+  # If we're last, don't print a trailing comma.
+  if [[ ${INDEX} -eq ${COUNT}-1 ]]; then
+    # Hack to output JSON.
+    cat << EOF >> ${JSON_FILE}
+    {
+	"name": "${TEST}",
+        "unit": "Seconds",
+	"value": ${TIME}
+    }
+EOF
+  else
+    cat << EOF >> ${JSON_FILE}
+    {
+	"name": "${TEST}",
+        "unit": "Seconds",
+	"value": ${TIME}
+    },
+EOF
+  fi
+
+  let INDEX=${INDEX}+1
+done
+
+cat << EOF >> ${JSON_FILE}
+]
+EOF
+
+cat ${JSON_FILE}
--- a/tests/run-ci-benchmarks.sh
+++ b/tests/run-ci-benchmarks.sh
@@ -1,10 +1,10 @@
-#!/bin/bash
+#!/usr/bin/env bash
+set -euo pipefail -o noclobber
+# set -xv # uncomment to debug variables
 
 JSON_FILE="benchmark-data.json"
 
-cat << EOF >> ${JSON_FILE}
-[
-EOF
+echo "[" >> "${JSON_FILE}"
 
 DIRNAME=$(dirname "$0")
 
@@ -16,13 +16,10 @@ for TEST in ${SUCC}; do
 done
 
 INDEX=0
-echo $SUCC
-echo $COUNT
 for TEST in ${SUCC}; do
-  echo ${INDEX}
 
   # Record wall clock time in seconds
-  /usr/bin/time -f "%e" -o /tmp/time cn verify ${TEST}
+  /usr/bin/time -f "%e" -o /tmp/time cn verify "${TEST}"
   TIME=$(cat /tmp/time)
 
   # If we're last, don't print a trailing comma.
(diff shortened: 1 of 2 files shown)


tests/run-ci-benchmarks.sh
```diff
@@ -0,0 +1,55 @@
+#!/bin/bash
```
dc-mak: Usually we start with
```bash
#!/usr/bin/env bash
set -euo pipefail -o noclobber
# set -xv # uncomment to debug variables
```

tests/run-ci-benchmarks.sh
```diff
@@ -0,0 +1,55 @@
...
+
+JSON_FILE="benchmark-data.json"
+
+cat << EOF >> ${JSON_FILE}
+[
+EOF
```
dc-mak:
```bash
echo "[" >> "${JSON_FILE}"
```

tests/run-ci-benchmarks.sh
```diff
@@ -0,0 +1,55 @@
...
+
+DIRNAME=$(dirname "$0")
+
+SUCC=$(find "${DIRNAME}"/cn -name '*.c' | grep -v '\.error\.c')
```
dc-mak:
```bash
SUCC=($(find "${DIRNAME}"/cn -name '*.c' | grep -v '\.error\.c'))
```
But measuring the time taken for failing benchmarks would be useful too right?

tests/run-ci-benchmarks.sh
```diff
@@ -0,0 +1,55 @@
...
+
+COUNT=0
+for TEST in ${SUCC}; do
+  let COUNT=${COUNT}+1
+done
```
dc-mak:
```bash
LEN=${#SUCC[@]}
```

tests/run-ci-benchmarks.sh
```diff
@@ -0,0 +1,55 @@
...
+
+INDEX=0
+echo $SUCC
+echo $COUNT
```
dc-mak: Is echoing the whole array necessary?

tests/run-ci-benchmarks.sh
```diff
@@ -0,0 +1,55 @@
...
+for TEST in ${SUCC}; do
```
dc-mak: You can iterate through the keys: http://stackoverflow.com/questions/3112687/ddg#3113285 so INDEX is unnecessary.

tests/run-ci-benchmarks.sh
```diff
@@ -0,0 +1,55 @@
...
+EOF
+  else
+    cat << EOF >> ${JSON_FILE}
+    {
+	"name": "${TEST}",
+        "unit": "Seconds",
+	"value": ${TIME}
+    },
+EOF
+  fi
```
dc-mak, line 43: Perhaps printf and a variable for the last comma would reduce code duplication here.

tests/run-ci-benchmarks.sh
```diff
@@ -0,0 +1,55 @@
...
+
+  let INDEX=${INDEX}+1
+done
+
+cat << EOF >> ${JSON_FILE}
+]
+EOF
```
dc-mak:
```bash
echo "]" >> "${JSON_FILE}"
```

tests/run-ci-benchmarks.sh
```diff
@@ -0,0 +1,55 @@
...
+
+cat ${JSON_FILE}
```
dc-mak: Optional tip: (Installing and) using `jq` will colorise (and format) the output.

dc-mak (commented):
If we envision this being updated and getting more complicated, then it would be worth considering a Python or OCaml script, but this is quite useful to have sooner than later so I'm willing to merge now and get the highlighted changes later.

Also consider using https://www.shellcheck.net/

tests/run-ci-benchmarks.sh
jprider63: Makes sense. I've added failing tests.

tests/run-ci-benchmarks.sh
jprider63: I spent a while attempting to get this to work without success. I don't think it's a proper array although I don't know bash that well. 

I'm inclined to leave the script as is for now and switch to a proper benchmarking system later.

dc-mak: That's fine (the issue is that there needs to be an extra pair of parens on the rhs of the defining TESTS).

jprider63:
Something is still broken with the failing tests. I'm working on debugging it.

jprider63:
@dc-mak Are you able to reopen this or should I create a new PR? I just pushed 3a1c0bb68a696a457f1bfe7b1ee1b5fa8c31458e to this branch, which should fix the issue. I'm currently running it on our fork [here](https://github.com/GaloisInc/cerberus/actions/runs/10585515969/job/29332260659) so we'll see if that passes.

jprider63:
FYI it [passed](https://github.com/GaloisInc/cerberus/actions/runs/10585515969/job/29332260659) on our fork.

dc-mak:
I'm not able to do anything on this PR, so a new one will be needed.
For future PRs, please mark as draft until ready.

PR #538, opened 2024-08-27
Revert "Add benchmarking CI that generates graphs on updates to master"

dc-mak:
Reverts rems-project/cerberus#530


dc-mak:
Unfortunately, because GitHub doesn't run new/updated workflows in PR, a failing workflow got merged in.

PR #542, opened 2024-08-28
[CN-Exec] Minor changes to runtime library to accomodate testing

ZippeyKeys12:
- include guards on all headers
- allowing the header files to be imported directly into C++ files
- being able to set the callback used by `cn_exit`.


rbanerjee20 (approved):
Looks good to me!

PR #547, opened 2024-08-28
[CN] Use only one global model evaluator solver

### LINKED ISSUE #369: [CN] the solver is using a lot of pipes
When running the CN ci tests (`tests/run-cn.sh`) on macOS (tested on 14.5) 10 tests fail because the `cn` process runs out of file descriptors (the macOS's default soft limit appears to be 256):

```
10 tests failed:
   ./cn/tree16/as_partial_map/tree16.c ./cn/tree16/as_mutual_dt/tree16.c ./cn/fun_ptr_three_opts.c ./cn/list_rev01.c ./cn/append.c ./cn/reverse.c ./cn/mutual_rec/mutual_rec.c ./cn/fun_ptr_known.c ./cn/mergesort.c ./cn/fun_ptr_extern.c
```

I think that the issue is a result of:
1. every call to `Simple_smt.new_solver` opens 3 pipes as a result of the call to `Unix.open_process_args_full`.
2. every call to `Solver.model_evaluator _ _` (called from `Solver.provable`) creates a new solver, whose lifetime extends until the end of the execution of CN (?).

dsainati1:
Closes https://github.com/rems-project/cerberus/issues/369


backend/cn/lib/solver.ml
```diff
@@ -1238,17 +1238,26 @@ let make globals =
   s
 
 
+let model_evaluator_solver = ref None
+
```
dc-mak: Nit: Unless this needs to be shared/used by other functions, then something like this is preferable.
```ocaml
(** Evaluate terms in the context of a model computed by the solver. *)
let model_evaluator =
  let let model_evaluator_solver = ref None in
  fun solver mo ->
  match SMT.to_list mo with
```

backend/cn/lib/solver.ml
```diff
@@ -1259,17 +1268,17 @@ let model_evaluator solver mo =
         globals = gs
       }
     in
-    declare_solver_basics evaluator;
-    List.iter (debug_ack_command evaluator) defs;
+    if new_solver then declare_solver_basics evaluator;
     fun e ->
       push evaluator;
+      List.iter (debug_ack_command evaluator) defs;
```
dc-mak: Why is this moving underneath `fun e ->`?

backend/cn/lib/solver.ml
```diff
@@ -1259,17 +1268,17 @@ let model_evaluator solver mo =
...
       let inp = translate_term evaluator e in
-      (match SMT.check s with
+      (match SMT.check evaluator.smt_solver with
        | SMT.Sat ->
-         let res = SMT.get_expr s inp in
+         let res = SMT.get_expr evaluator.smt_solver inp in
          pop evaluator 1;
          let ctys = get_ctype_table evaluator in
-         Some (get_ivalue gs ctys (basetype e) (SMT.no_let res))
+         Some (get_ivalue solver.globals ctys (basetype e) (SMT.no_let res))
```
dc-mak: Any reason for this rename?

backend/cn/lib/solver.ml
```diff
@@ -1259,17 +1268,17 @@ let model_evaluator solver mo =
...
     in
-    declare_solver_basics evaluator;
-    List.iter (debug_ack_command evaluator) defs;
+    if new_solver then declare_solver_basics evaluator;
     fun e ->
       push evaluator;
+      List.iter (debug_ack_command evaluator) defs;
       let inp = translate_term evaluator e in
-      (match SMT.check s with
+      (match SMT.check evaluator.smt_solver with
```
dc-mak: Any reason for `evaluator.smt_solver` as opposed to `smt_solver`?

backend/cn/lib/solver.ml
```diff
@@ -1259,17 +1268,17 @@ let model_evaluator solver mo =
...
        | SMT.Sat ->
-         let res = SMT.get_expr s inp in
+         let res = SMT.get_expr evaluator.smt_solver inp in
```
dc-mak: Any reason for `evaluator.smt_solver` as opposed to `smt_solver`?

dc-mak (commented):
Thanks! Some minor nits and clarifications.

backend/cn/lib/solver.ml
dsainati1: This delays the loading of model definitions into the solver until the model is actually used, as opposed to now wherein we can create a solver and load it with definitions and never actually query it for anything.

backend/cn/lib/solver.ml
dsainati1: This and below are leftovers from a previous approach, will revert.

backend/cn/lib/solver.ml
dc-mak: But does that not repeat the re-definitions every time the solver is used subsequently?

dsainati1: Ah yes, good call. I added a guard for this.

dsainati1: Hmm yeah, good point. We cannot load the definitions when the solver is created anymore because it's only created once, so we need to change the push/pop behavior to only pop when the requested model is different from the one we currently have. Will look into this.

yav (approved):
LGTM

PR #552, opened 2024-09-02
[CN-exec] Add `instrument` subcommand

### LINKED ISSUE #480: [CN] Add subcommand for executable spec
Continuation of https://github.com/rems-project/cerberus/issues/417

rbanerjee20:
Add `instrument` subcommand for CN executable spec generation. CN verification mode has been separated from executable spec mode in the following way:

* Providing values for any of the `Executable_spec_flags` no longer affects the `verify` subcommand in any way
* Calling `cn instrument <FILE>` without any of the `Executable_spec_flags` defined should now work, whereas before the `--output-decorated` flag needed to be set with some output filename for executable specs to be generated at all

Fixes https://github.com/rems-project/cerberus/issues/480

### DIFF
--- a/backend/cn/bin/main.ml
+++ b/backend/cn/bin/main.ml
@@ -346,8 +346,7 @@ let verify
       Typing.run_from_pause check paused)
 
 
-let generate_executable_specs 
-  (* Common *)
+let generate_executable_specs
   filename
   macros
   incl_dirs
@@ -372,12 +371,12 @@ let generate_executable_specs
   no_inherit_loc
   magic_comment_char_dollar
   (* Executable spec *)
-  output_decorated
+    output_decorated
   output_decorated_dir
   with_ownership_checking
   with_test_gen
   copy_source_dir
-  = 
+  =
   if json then (
     if debug_level > 0 then
       CF.Pp_errors.fatal "debug level must be 0 for json output";
@@ -906,7 +905,8 @@ let runtime_testing_cmd =
     $ Executable_spec_flags.copy_source_dir
   in
   let doc =
-    "Instruments [FILE] with runtime C assertions that check the properties provided in CN specifications.\n"
+    "Instruments [FILE] with runtime C assertions that check the properties provided in \
+     CN specifications.\n"
   in
   let info = Cmd.info "runtime-test" ~doc in
   Cmd.v info runtime_testing_t
--- a/backend/cn/bin/main.ml
+++ b/backend/cn/bin/main.ml
@@ -868,9 +868,9 @@ let generate_tests_cmd =
   Cmd.v info generate_tests_t
 
 
-let runtime_test_cmd =
+let instrument_cmd =
   let open Term in
-  let runtime_test_t =
+  let instrument_t =
     const generate_executable_specs
     $ Common_flags.file
     $ Common_flags.macros
@@ -905,11 +905,11 @@ let runtime_test_cmd =
     "Instruments [FILE] with runtime C assertions that check the properties provided in \
      CN specifications.\n"
   in
-  let info = Cmd.info "runtime-test" ~doc in
-  Cmd.v info runtime_test_t
+  let info = Cmd.info "instrument" ~doc in
+  Cmd.v info instrument_t
 
 
-let subcommands = [ wf_cmd; verify_cmd; generate_tests_cmd; runtime_test_cmd ]
+let subcommands = [ wf_cmd; verify_cmd; generate_tests_cmd; instrument_cmd ]
 
 let () =
   let version_str = Cn_version.git_version ^ " [" ^ Cn_version.git_version_date ^ "]" in
--- a/backend/cn/bin/main.ml
+++ b/backend/cn/bin/main.ml
@@ -868,9 +868,9 @@ let generate_tests_cmd =
   Cmd.v info generate_tests_t
 
 
-let runtime_testing_cmd =
+let runtime_test_cmd =
   let open Term in
-  let runtime_testing_t =
+  let runtime_test_t =
     const generate_executable_specs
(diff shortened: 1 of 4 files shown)


backend/cn/lib/executable_spec.ml
```diff
@@ -199,11 +199,21 @@ let main
...
   statement_locs
   =
+  let output_filename =
+    match output_decorated with
+    | None ->
+      let strs = String.split_on_char '/' filename in
+      let last_str = List.nth strs (List.length strs - 1) in
+      let last_str_basename_list = String.split_on_char '.' last_str in
+      let last_str_basename = List.nth last_str_basename_list 0 in
+      last_str_basename ^ "-exec.c"
```
dc-mak: https://ocaml.org/manual/5.2/api/Filename.html specifically `basename` I guess?

backend/cn/bin/main.ml
```diff
@@ -819,7 +868,48 @@ let generate_tests_cmd =
...
+  in
+  let doc =
+    "Instruments [FILE] with runtime C assertions that check the properties provided in \
+     CN specifications.\n"
+  in
+  let info = Cmd.info "runtime-test" ~doc in
+  Cmd.v info runtime_testing_t
+
+
+let subcommands = [ wf_cmd; verify_cmd; generate_tests_cmd; runtime_testing_cmd ]
```
dc-mak: Nit: seems a bit inconsistent.

| subcommand | variable name |
|---------------|------------------|
| `wf` | `wf_cmd` |
| `verify` | `verify_cmd` |
| `generate-tests`  | `generate_tests_cmd` |
| `runtime-test` | `runtime_testing_cmd` |

dc-mak (commented):
Thanks! I guess the intention is to add the subcommand now and then prune the flags later? Any reason for not pruning irrelevant flags in one go?

rbanerjee20:
> Thanks! I guess the intention is to add the subcommand now and then prune the flags later? Any reason for not pruning irrelevant flags in one go?

When you say prune irrelevant flags, what do you mean?

backend/cn/lib/executable_spec.ml
rbanerjee20: Thanks!

dc-mak:
> > Thanks! I guess the intention is to add the subcommand now and then prune the flags later? Any reason for not pruning irrelevant flags in one go?
> 
> When you say prune irrelevant flags, what do you mean?

It looks like the `runtime-test` subcommand uses more than just the `Executable_spec_flags` or `Common` flags and I was wondering why. For example, the `use_vip` flag - does that need to move to `Common`? The `json` flag, is that relevant to `runtime-test`?

backend/cn/lib/executable_spec.ml
```diff
@@ -206,12 +206,7 @@ let main
   =
   let output_filename =
     match output_decorated with
-    | None ->
-      let strs = String.split_on_char '/' filename in
-      let last_str = List.nth strs (List.length strs - 1) in
-      let last_str_basename_list = String.split_on_char '.' last_str in
-      let last_str_basename = List.nth last_str_basename_list 0 in
-      last_str_basename ^ "-exec.c"
+    | None -> Filename.remove_extension (Filename.basename filename) ^ "-exec.c"
```
dc-mak: Nit: You can open a module by doing  `ModuleName.(<expr>)`  e.g.
```ocaml
Filename.(remove_extension (basename filename))
```

ZippeyKeys12:
I think `runtime-test` suggests this command runs tests, but this is only the instrumentation step, right?

rbanerjee20:
> I think `runtime-test` suggests this command runs tests, but this is only the instrumentation step, right?

True, we discussed this internally as well... it's a bit of a difficult one, maintaining brevity but also accuracy of the subcommand name. `runtime-instrument`?

@ZippeyKeys12 @dc-mak let me know what you think.

rbanerjee20:
> > > Thanks! I guess the intention is to add the subcommand now and then prune the flags later? Any reason for not pruning irrelevant flags in one go?
> > 
> > 
> > When you say prune irrelevant flags, what do you mean?
> 
> It looks like the `runtime-test` subcommand uses more than just the `Executable_spec_flags` or `Common` flags and I was wondering why. For example, the `use_vip` flag - does that need to move to `Common`? The `json` flag, is that relevant to `runtime-test`?

Ah, understood. I am not sure actually, since I copied these across from the function called for the `verify` subcommand and only deleted the solver-specific flags since I know those definitely aren't relevant to executable-spec generation. Perhaps we can leave the pruning to a future issue/PR after we have discussed which flags are relevant to runtime testing and which aren't?

dc-mak:
fair enough, as long as there's a todo or an issue somewhere

rbanerjee20:
Sure, I’ll make one @dc-mak

ZippeyKeys12:
> True, we discussed this internally as well... it's a bit of a difficult one, maintaining brevity but also accuracy of the subcommand name. `runtime-instrument`?

Is just `instrument` not accurate? Otherwise, `runtime-instrument` makes sense, it's just a bit longer.

rbanerjee20:
> > True, we discussed this internally as well... it's a bit of a difficult one, maintaining brevity but also accuracy of the subcommand name. `runtime-instrument`?
> 
> Is just `instrument` not accurate? Otherwise, `runtime-instrument` makes sense, it's just a bit longer.

`instrument` is accurate, it just seemed a bit vague to me. We already had a discussion about this a few weeks ago internally but I'll ask around again tomorrow; this sort of thing seems important for the user experience (even though it's very easy to change the subcommand name later if we change our minds).

rbanerjee20:
Let's just go with `instrument` :)

PR #554, opened 2024-09-03
CN VIP: Add support for ptr diff

dc-mak:
This commit adds support for the VIP ptr diff rule, which requires that both pointers come from the same live allocation.

To accommodate this without excessive annotation burden, it does so by (a) inferring `base <= (u64) p <= (u64) p + sizeof(ct) <= base + len` for `{ base, len } = allocs[(alloc_id)p]` from `Owned<ct>(p)` and (b) accepting an `Owned<ct>(p)` when looking for a `Alloc(q)` if `(alloc_id) q == (alloc_id) p` and `(u64) p <= (u64) q <= (u64) p + sizeof(ct)`.

It only does this for pointer difference for now (later, it will be used for pointer relational operators, copy_alloc_id and integer to pointer round-trip casts) with a special flag `alloc_or_owned` in the resource inference algorithm.

### DIFF
--- a/backend/cn/lib/check.ml
+++ b/backend/cn/lib/check.ml
@@ -1308,15 +1308,45 @@ let rec check_expr labels (e : BT.t mu_expr) (k : IT.t -> unit m) : unit m =
                 | ct -> Memory.size_of_ctype ct
               in
               let both_alloc =
-                and_ [ hasAllocId_ arg1 here; hasAllocId_ arg2 here ] here
+                and_
+                  [ hasAllocId_ arg1 here;
+                    hasAllocId_ arg2 here;
+                    eq_ (allocId_ arg1 here, allocId_ arg2 here) here
+                  ]
+                  here
               in
+              let ub = CF.Undefined.(UB_CERB004_unspecified UB_unspec_pointer_sub) in
               let@ provable = provable loc in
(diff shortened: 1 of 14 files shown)


backend/cn/lib/check.ml
```diff
@@ -1308,15 +1308,45 @@ let rec check_expr labels (e : BT.t mu_expr) (k : IT.t -> unit m) : unit m =
...
+              let ub = CF.Undefined.(UB_CERB004_unspecified UB_unspec_pointer_sub) in
               let@ provable = provable loc in
               match provable @@ t_ @@ both_alloc with
               | `False ->
                 let@ model = model () in
-                let ub = CF.Undefined.(UB_CERB004_unspecified UB_unspec_pointer_sub) in
                 fail (fun ctxt -> { loc; msg = Undefined_behaviour { ub; ctxt; model } })
               | `True ->
+                let@ (_alloc_or_owned, RE.O base_len), _ =
+                  RI.Special.is_pointer_live loc Ptr_diff arg1
```
dc-mak, line 1326: More accurate: `is_alloc_live`

dc-mak (commented):
* Fix naming of resource inference
* Name test functions more descriptively
* Add test for in pointer live _inside_ owned footprint
* Refactor resource inference?

PR #556, opened 2024-09-04
CN VIP: Generalise ptr diff support

dc-mak:


tests/cn/has_alloc_id_array_shift.c
dc-mak, line 1: This should include a test for member shifting too

PR #569, opened 2024-09-11
Interval analysis to simplify resource type permissions.

### LINKED ISSUE #351: [CN] simplify guards in `each` terms in state output
I got a program with this term in the Available resources: `each(u64 j; (((j < 4u64) && (!(j == 0u64))) && (!(true && (j < 2u64)))) && (!(j == 2u64))) {Block<struct instrumentation_state>(&instrumentation + j * 44u64 /* 0x2c */)}(ii) (same type)`

I simplified `(((j < 4u64) && (!(j == 0u64))) && (!(true && (j < 2u64)))) && (!(j == 2u64))` to `((j < 4u64) && (j >= 3u64))` manually. This is much easier to read. I might have made a mistake.

Simplifying terms in general might be tricky but at least
`(!(true && (j < 2u64)))` should be amenable to local simplification to `(j >= 2u64)`.

### LINKED ISSUE #389: [CN] Simplified Constraints in state file are not fully simplified
I struggled to find a minimal model to show this, so I'll just show a picture of the output I'm getting.

Screenshot 2024-07-16 at 10 16 10 AM

It's not important to understand, but the point is that there are extremely long terms in this constraint which do not have to be so long. Under `terms` it shows what some sub terms simplify to, so I manually have to go in and substitute those simplifications in myself. This work is very tedious and seems like it can be automated.

To try to make things a bit clearer with this example, you can note that the unproven constraint includes the term `unpack_Own_Forwards0.n.data`. Under terms, it can be seen that `unpack_Own_Forwards0.n.data` has the value `0i32`. I can't see a good reason why the unsimplified constraint needs this long term that could just be simplified to `0i32`.

Sorry that this is a bit confusing to understand, I'm happy to try to clarify anything.

yav:
Improves the situation with #389, fixes #351

This modifies `explain.ml` to call a function on quantified resource predicates, which simplifies their permission fields.
The idea is to compute the intervals based on the current model and present them in a normalized form.

For example, previously we would print:
```
each(u64 i; 0'u64 <= i && i < 48'u64 /* 0x30 */ && i != 0'u64 && !(true && 0'u64 <= i && i < 32'u64 /* 0x20 */) && i != 32'u64 /* 0x20 */)
```
after simplification we get:
```
each(u64 i; 1'u64 <= i && i < 32'u64 /* 0x20 */ || 33'u64 /* 0x21 */ <= i && i < 48'u64 /* 0x30 */)
```

### DIFF
--- a/backend/cn/lib/explain.ml
+++ b/backend/cn/lib/explain.ml
@@ -132,7 +132,9 @@ let rec simp_resource eval r =
     else
       LAT.Constraint (ct, info, simp_resource eval more)
   | LAT.Define (x, i, more) -> LAT.Define (x, i, simp_resource eval more)
-  | LAT.Resource (x, i, more) -> LAT.Resource (x, i, simp_resource eval more)
+  | LAT.Resource ((x,(rt,bt)), i, more) ->
+    let rt1 = Interval.Solver.simp_rt eval rt in
+    LAT.Resource ((x,(rt1,bt)), i, simp_resource eval more)
   | I i -> I i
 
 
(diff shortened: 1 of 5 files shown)


cp526:
Thanks, that looks very useful! 

Do I understand correctly that this always prints resources as specialised to the counterexample values? We'll want to be able to see both views, the fully concrete one with your simplification, but also the symbolic one we have currently. Perhaps we could put the two views (symbolic/concretised) side-by-side?

Aside from that, it looks like your simplification is only applied to the "interesting" resources, not the uninteresting ones.

yav:
@cp526 I've just made the suggested fixes---for constraints and resources that were simplified, we get a little toggle button that you can press to switch between the simplified and original views.  The simplifications should now be applied to both the interesting and non interesting categories.

cp526:
That's great, thanks very much!

PR #581, opened 2024-09-18
[CN-Exec] Add `convert_from_cn_map`

ZippeyKeys12:
You can't assign directly to an array, so it has to be a statement.
This means there's special-casing in `cn_internal_to_ail.ml`


rbanerjee20 (approved):
LGTM!

Issue #539, opened 2024-08-27
[CN] countermodels don't always satisify requires constraints/predicates/definitions


cassiatorczon:
identify constraints, predicates, and terms not passed to solver and mark constraints known to not be satisfied by countermodel

PR #582, opened 2024-09-18
display constraints and definitions not given to solver in error file

cassiatorczon:
Part of #539
Display constraints, predicate definitions, and function definitions that are not given to the solver in the error file.
Separates them into interesting/uninteresting.


cp526:
This looks very good. 

One minor point: in the output html, the "Definitions and constraints not given to solver" is phrased in implementation terms rather than what this means to the user. Maybe "Definitions and constraints not handled fully automatically." with "These may require manual proof steps and may not be correctly reflected in counterexamples." as an additional tooltip hint?

Aside from that, one easy extensions: we might want to also list (maybe in the same list, maybe separately) uninterpreted CN functions and resource predicates.

cp526:
Regarding the pull request, I'm seeing "This branch cannot be rebased due to conflicts", so we'd have to go with "squash and merge".

cassiatorczon:
- Have we decided on a central documentation to maintain that could be linked in the tooltip? If I were a user not familiar with the implementation, my next question on seeing that a constraint is not correctly reflected in counterexamples would be why isn't it and how do I change it so that it is.
- It should already be showing uninterpreted CN functions and resource predicates (the "definitions" in "definitions and constraints"). Is there a different name that would make that clearer?
- Where does it say that it can't be rebased? I'm only seeing something that says "This branch has no conflicts with the base branch", and it's giving me "rebase and merge" as an option. Happy to resolve conflicts, but I'm not sure what they are.

cp526:
> * Have we decided on a central documentation to maintain that could be linked in the tooltip? If I were a user not familiar with the implementation, my next question on seeing that a constraint is not correctly reflected in counterexamples would be why isn't it and how do I change it so that it is.

Good point. We don't have that in place yet. Indeed we should probably have a section of text dedicated to just this situation and then link to that. But in the mean time, probably still useful to rephrase a bit.

> * It should already be showing uninterpreted CN functions and resource predicates (the "definitions" in "definitions and constraints"). Is there a different name that would make that clearer?

Mm, maybe I'm not following correctly. Doesn't `not_given_to_solver` in `Context` only return recursive functions and multi-branch resource predicates (via `ResourcesPredicates.given_to_solver`)?

> * Where does it say that it can't be rebased? I'm only seeing something that says "This branch has no conflicts with the base branch", and it's giving me "rebase and merge" as an option. Happy to resolve conflicts, but I'm not sure what they are.

That's odd. It's showing this for me at the bottom of this page (https://github.com/rems-project/cerberus/pull/582) : "All checks have passed. This branch cannot be rebased due to conflicts." No idea where the difference in views would be coming from. Don't know if there's a way to view the conflicts in the github interface (not a github expert).

cassiatorczon:
Oh, I misread - you're right about the uninterpreted functions and resource predicates. Do you know where I would get that information for predicates?

cp526:
> Oh, I misread - you're right about the uninterpreted functions and resource predicates. Do you know where I would get that information for predicates?

The `clauses` field of a `ResourcePredicates.definition` is `None` in the uninterpreted case.

cp526:
That looks great!

cp526:
I'm still getting "This branch cannot be rebased due to conflicts" (for whatever reason), so I'll have to go with 'squash and merge'.

bcpierce00:
Cool.  Can somebody post a screenshot of the new display?

PR #585, opened 2024-09-20
[CN-exec] Fix anonymous records

rbanerjee20:
Previously, anonymous records only properly worked in CN runtime testing when they were the return type of some CN predicate or function. Now anonymous records should work across the board at runtime, getting translated to C structs as before.

~~TODO: Revert CN tutorial examples to use records instead of datatypes and check that it works now.~~
~~TODO: Write small examples of anonymous records that are not the return types of CN predicates/functions and check that they work. Include in CI.~~


backend/cn/lib/cn_internal_to_ail.ml
```diff
@@ -54,37 +54,15 @@ module MembersKey = struct
...
+    | (id, cn_bt) :: ms, (id', cn_bt') :: ms' ->
       let c = String.compare (Id.s id) (Id.s id') in
-      if c == 0 then
-        0
-      else
+      let c' = BaseTypes.compare (cn_base_type_to_bt cn_bt) (cn_base_type_to_bt cn_bt') in
+      if c == 0 && c' == 0 then
         compare ms ms'
+      else
+        c
```
dc-mak, line 68: I think something like this? if the point is lexicographic, then this avoids unnecessary computation
```ocaml
let c = .. in 
if c == 0 then
  let c' = ... in
  if c' == 0 then
    recurse
  else
    c'
else
  c
```

backend/cn/lib/cn_internal_to_ail.ml
```diff
@@ -236,11 +204,12 @@ let rec cn_to_ail_base_type ?(pred_sym = None) cn_typ =
       generate_ail_array bt
       (* TODO: What is the optional second pair element for? Have just put None for now *)
     | CN_tuple ts ->
+      failwith "TODO"
```
dc-mak: `failwith (__FUNCTION__  ": more detailed explanation")` is usually more helpful to non-devs/users

backend/cn/lib/cn_internal_to_ail.ml
```diff
@@ -2922,6 +2891,7 @@ let cn_to_ail_function_internal
       in
       (bs, Some (List.map mk_stmt ss))
     | Uninterp -> ([], None)
+    (* TODO: Make this throw a clear error. Uninterpreted functions must not exist or be used for runtime checking to work *)
```
dc-mak: `failwith`? is this likely to be confusing?

backend/cn/lib/executable_spec_records.ml
```diff
@@ -0,0 +1,139 @@
...
+  | WrapI (_ity, t) -> add_records_to_map_from_it t
+  | Aligned { t; align } -> List.iter add_records_to_map_from_it [ t; align ]
+  | MapConst (_bt, t) -> add_records_to_map_from_it t
+  | MapSet (t1, t2, t3) -> List.iter add_records_to_map_from_it [ t1; t2; t3 ]
+  | MapGet (t1, t2) -> List.iter add_records_to_map_from_it [ t1; t2 ]
+  | MapDef ((_, _), t) -> add_records_to_map_from_it t
+  | Apply (_pred, ts) -> List.iter add_records_to_map_from_it ts
+  | Let ((_, t1), t2) -> List.iter add_records_to_map_from_it [ t1; t2 ]
+  | Match (e, cases) -> List.iter add_records_to_map_from_it (e :: List.map snd cases)
+  | Constructor (_sym, args) -> List.iter add_records_to_map_from_it (List.map snd args)
```
dc-mak, line 55: `IT.fold_` or `IT.fold_subterms`?

backend/cn/lib/cn_internal_to_ail.ml
```diff
@@ -186,26 +159,21 @@ let str_of_bt_bitvector_type sign size =
...
-       sym'
-     | (_, sym') :: _ -> sym')
+let augment_record_map ?cn_sym bt =
+  let sym_prefix = match cn_sym with Some sym' -> sym' | None -> Sym.fresh () in
+  match bt_to_cn_base_type bt with
+  | CN_record members ->
+    (* Augment records map if entry does not exist already *)
+    if not (RecordMap.mem members !records) then (
+      let sym' = generate_sym_with_suffix ~suffix:"_record" sym_prefix in
+      records := RecordMap.add members sym' !records)
```
dc-mak, line 174: Might be better to use a Hashtable if you want/need mutability? But do you need mutability?

dc-mak (commented):
Few nits/questions, but looks reasonable overall

backend/cn/lib/cn_internal_to_ail.ml
rbanerjee20: Shouldn't need mutability - a record type is only ever added to the `RecordMap` in this initial pass through the CN spec and then possibly accessed at some later point during the translation of a record type or at the point of instantiation of an anonymous record.

backend/cn/lib/executable_spec_records.ml
rbanerjee20: Thanks, I had a look but I am unsure how to encode what I want to do using (e.g.) `IT.fold_`. Specifically, the `RecordMap` needs to be augmented every time we reach a record, which I have written like this:

```ml
  | ...
  | Record members ->
    (* Anonymous record instantiation -> add to records map *)
    Cn_internal_to_ail.augment_record_map (IT.bt it);
    List.iter (fun (_, it') -> add_records_to_map_from_it it') members
  | ...
```

 but the `Record` case of `IT.fold_` does this:

```ml
  | ...
  | Record members -> fold_list f binders acc (List.map snd members)
  | ...
```

so I can't really intercept this case with the necessary map operation in the same way I think. `IT.fold_subterms` looks similar, just with an optional `bindings` argument and then a call to `fold`.

PR #537, opened 2024-08-27
[CN] Add benchmarking for PRs

jprider63:
This PR adds a CI workflow that adds regression benchmarks to PRs. It builds upon #530 so that should be merged first (ignore the first three commits). 

Every time a PR is updated, a comment is posted to the PR that displays the old runtimes and the new runtimes. It highlights any regressions and if any benchmarks were removed. An example comment is [here](https://github.com/GaloisInc/cerberus/pull/5#issuecomment-2312896407).

Implements part of https://github.com/rems-project/cn-tutorial/issues/54.

PR #597, opened 2024-09-25
Benchmark z3 and cvc5

jprider63:
This updates the benchmarking CI to display results for z3 and cvc5. An example of what this looks like is available [here](https://galoisinc.github.io/cerberus/dev/bench/).

**Note**: #596 must be merged first.

### DIFF
--- a/.github/workflows/ci-bench.yml
+++ b/.github/workflows/ci-bench.yml
@@ -94,19 +94,23 @@ jobs:
         opam pin --yes --no-action add cn .
         opam install --yes cn ocamlformat.0.26.2
 
-    - name: Run benchmark
+    - name: Run benchmarks
       run: |
         opam switch ${{ matrix.version }}
         eval $(opam env --switch=${{ matrix.version }})
-        cd tests; USE_OPAM='' ./run-ci-benchmarks.sh
+        cd tests; SOLVER='z3' ./run-ci-benchmarks.sh; SOLVER='cvc5' ./run-ci-benchmarks.sh
         cd ..
 
     - name: Store benchmark result
-      uses: benchmark-action/github-action-benchmark@v1
+      uses: GaloisInc/github-action-benchmark@00bf34008c7b91a09ea7f1ef93330ccb7bf2d4ab
       with:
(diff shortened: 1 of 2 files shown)


jprider63:
I found one edge case. I'm fixing it now.

jprider63:
This is now fixed!

PeterSewell:
Looks good.

I wouldn't want to suggest anyone tinkers endlessly with the visualisation
- but on the other hand it'd be lovely to be able to see the comparisons in
a slightly more granular way.  Perhaps it'd be easy to fix an order of the
tests, eg order by increasing Z3 runtime in the current version, and then
draw bargraphs with the tests on the x axis and four thin bars (z3
previous, z3 current, cvc5 previous, cvs current) for each test?

Is buddy in the CI?

p

cp526:
> Is buddy in the CI? 

Not currently. Last we checked the performance was bad enough we couldn't tell whether it's even in a working state, so we can't include it in CI at the moment.

jprider63:
> Perhaps it'd be easy to fix an order of the tests, eg order by increasing Z3 runtime in the current version

The current CI visualization tool does not support ordering. I can extend it, but it would take some time. 

> draw bargraphs with the tests on the x axis and four thin bars (z3 previous, z3 current, cvc5 previous, cvs current) for each test

[These graphs](https://rems-project.github.io/cerberus/dev/bench/) currently show how the performance changes over time. They display runtimes with z3 and cvc5 for every update to `master`, where the x-axis is the update's commit. Would you prefer a single bar graph that only displays the last two updates instead? This would lose the history over time information that would help us identify future regressions. 

I've also implemented this [PR](https://github.com/rems-project/cerberus/pull/537) which runs regression benchmarks on every pull request that is opened. It posts a comment on the PR that compares the previous performance with the PR's performance and highlights any regressions. An example on Galois' fork is [here](https://github.com/GaloisInc/cerberus/pull/5#issuecomment-2315287117). Unfortunately, it is currently not working on this repository due to Github permission errors that we haven't successfully debugged yet.

PR #601, opened 2024-09-30
[CN-Exec] Add logging levels to runtime library

ZippeyKeys12:
In the case of random testing, I don't think we want all of these logs, except for a failing input.

Instead I think we want to:
- disable logging
- test many random inputs
- if we find a failing one
  - re-enable logging at the error level
  - rerun the failing input
- we allow the user to re-run it again with the full logs


rbanerjee20:
This looks good, but I think we do want a debug level where we can see all the print statements as we are currently seeing them, which we would probably want to be configurable as a CLI option. I think this would be quite a bit more work though, since connecting up parts of the runtime library with the generated output seems less straightforward than just modifying the runtime library directly

ZippeyKeys12:
I'm not sure I understand, the default logging level `CN_LOGGING_INFO` shows all the current print statements. Do you mean also exposing the ability to change the logging level via the `cn instrument` command? If so, that should be easy to do, I can add it.

PR #604, opened 2024-10-02
remove z3 dependency in backend_common

giltho:
Hello,
Since `backend/common/smt.ml` is not used at the moment, there is no need for z3 to be a dependency of `backend_common` or of cerberus-lib.
Since z3 is quite a heavy dependency, I removed it for my "personal" usage, and thought I might as well propose a PR, but feel free to reject it, especially if you plan on using smt.ml again


giltho:
Oh one thing I forgot to mention: we did an experiment where we removed the z3 dependency, and started to talk to z3 through smtlib instead using [GaloisInc/simple-smt-ocaml](https://github.com/GaloisInc/simple-smt-ocaml) -- or a [simple fork of ours that can be installed](https://github.com/natkarmios/simple-smt-ocaml).
In addition to not requiring a build of z3 every other day, it turned out to be... faster. By like 2%, but still, it removed any reason to use the native API

PR #610, opened 2024-10-04
CN VIP: Implement byte rep frontend

dc-mak:
These commits implement the frontend plumbing for the `to_bytes` and `from_bytes` CN statements required to implement converting values to the their byte-level representation.

Though `to_bytes` (at this stage) only makes sense with `Owned/Block` and `from_bytes` with `each (..) { Owned/Block(..) };`, the parser is more general, and this check is enforced in `check.ml` (with accompanying tests).

There's also a clean-up of `cn_internal_to_ail` whilst I was in the area.


backend/cn/lib/cn_internal_to_ail.ml
dc-mak, line 1: @rbanerjee20 can you take a quick look over this file just to make sure the only changes are adding a bunch of `_` in a bunch of binders?

rbanerjee20: looks good to me!

PR #612, opened 2024-10-07
CN VIP: Add extra test for to-bytes dummy impl

dc-mak:


dc-mak:
Test was an older version of a file that's already been committed.

Issue #256, opened 2024-05-11
[CN] Hard to figure out that `extract` needs `u64` argument for array-typed blocks


septract:
When writing `extract` for array-typed blocks, the operator takes a `u64` value. This is hard to figure out, especially since the C-level index is an `int` / `i32`. 

Example: 
```
void
array_4()
{
    int arr[1];
    /*@ extract Block<int>, 0u64; @*/ // <-- non-obvious 
    arr[0] = 1;
}
```

See also the discussion with @cp526 on MM [here](https://mattermost.galois.com/verse-ta1-ta2/pl/g5e7xy5knib5f85euqx4iq7isc). Quoting the relevant part: 

> @cp526: "With CN as-is the `extract` needs a `u64` index, because a `Block<int[1]>` (or any other array-typed block) decomposes into an iterated resource indexed by `u64`. [...] That's not great UI-wise (having to match that so-far-invisible u64 index type), so I have to see what we do about that. Perhaps if we adopt your other suggestion (from a while back) of having `extract` additionally take the base pointer of the resource the extract should be applied to (which we should just do anyway), CN could provide some useful output about why nothing was extracted.

septract:
You need the same `u64` type for asserting properties of array offsets, eg the following (which is example 00090 in the `c-testsuite` dataset) 
```
int a[3] = {0, 1, 2};

int
main()
/*@ accesses a; @*/
/*@ requires 
			a[0u64] == 0i32; 
			a[1u64] == 1i32; 
			a[2u64] == 2i32; 
			@*/
/*@ ensures return == 0i32; @*/
{
	/*@ extract Owned<int>, 0u64; @*/
	/*@ extract Owned<int>, 1u64; @*/
	/*@ extract Owned<int>, 2u64; @*/
	if (a[0] != 0)
		return 1;
	if (a[1] != 1)
		return 2;
	if (a[2] != 2)
		return 3;
	
	return 0;
}
```

This is also not very intuitive.

peterohanley:
If your `each` statements do not have `u64` indices, they will also not be used to satisfy resources.

septract:
This issue showed up again in an example from BAE / @lwli11

jprider63:
I've started looking into this. I'm somewhat confused by what's going on. Examples like the following fail as described:
```
void
array_4()
{
    int arr[1];
    /*@ extract Block<int>, 0i32; @*/
    arr[0] = 1;
}
```

[...]

PR #616, opened 2024-10-08
Warn when a `u64` is not used for index types. 

jprider63:
This address #256 by throwing a warning when anything besides a `u64` is used for index types on `extract` and `array_shift`. I left this as a warning instead of an error for now since this would break existing code (like those in the tutorial). Here's an example of the output warning:
```
tests/cn/array_readex2.c:3:77: warning: 'array_shift' expects a 'u64', but 'j' with type 'i32' was provided. This will become an error in the future.
        requires take a1 = each(i32 j; 0i32 <= j && j < 1i32) { Owned<char>(array_shift<char>(tmp,j)) };
```
from this example:
```
void readex(char* tmp)
/*@
        requires take a1 = each(i32 j; 0i32 <= j && j < 1i32) { Owned<char>(array_shift<char>(tmp,j)) };
@*/
{
        return;
}
int tmp()
{
        char tmp[2]={'H', 'i'};
        readex(tmp);
        return 1;
}
```

Are there any other primitives besides `extract` and `array_shift` that this warning should be added to?

### DIFF
--- a/backend/cn/lib/compile.ml
+++ b/backend/cn/lib/compile.ml
@@ -773,9 +773,10 @@ module EffectfulTranslation = struct
                        pointer"
               }
         in
+        let@ index = self index in
+        Typing.warn_when_not_u64 "array_shift" loc (BaseTypes.Surface.proj (IT.bt index)) (Some index);
         (match IT.bt base with
          | Loc _ ->
-           let@ index = self index in
            (match IT.bt index with
             | Integer | Bits _ ->
               return (IT (ArrayShift { base; ct; index }, Loc (Some ct), loc))
@@ -868,7 +869,6 @@ module EffectfulTranslation = struct
             e
         in
         let@ bt = check_quantified_base_type env loc bt in
-        warn_when_not_u64 loc bt;
         return
           (IT
              ( EachI ((Z.to_int (fst r), (sym, SBT.proj bt), Z.to_int (snd r)), expr),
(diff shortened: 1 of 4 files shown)


cp526:
Thanks! Two minor points:
- I think we should move the call for those warnings into `wellTyped` (that's where other checks of the same kind are done)
- instead of the array-shift, we should put the check on the quantifier for iterated resources (`each`). Otherwise one could still run into the confusing situation (where an `extract` has no effect due to basetype mismatches) by quantifying `q` over a non-u64 basetype, but casting `q` to `u64` inside the array-shift.

The warnings would then move to `https://github.com/cp526/cerberus/blob/bea7d0034b850de731af40e1f4b1a411499d4737/backend/cn/lib/wellTyped.ml#L986` and `https://github.com/cp526/cerberus/blob/bea7d0034b850de731af40e1f4b1a411499d4737/backend/cn/lib/wellTyped.ml#L1824`.

jprider63:
This should be addressed now! Let me know if you have any other feedback.

cp526:
That looks great, thanks!

dc-mak:
I don't understand the nature of this change, or why we would want to forbid signed index types? I agree there should be a warning for when index types are mismatched but this seems to be heading in quite a restrictive direction?

https://github.com/rems-project/cerberus/pull/616/files#diff-cb79e893729b5a0fe3e9a4d831dc18887ca0a6c7b8d82a510079f61ca794272cR982

jprider63:
That sounds like a potentially better approach long term. From discussion [here](https://github.com/rems-project/cerberus/issues/256#issuecomment-2389242723), one thing that prevents this is that "`extract` does not uniquely identify a paired resource". This would be good to address in the syntax update discussions.

PR #626, opened 2024-10-11
[CN-Test-Gen] CI updates (new tests, failing cases, better readability)

ZippeyKeys12:
Includes the tests from #441, just doesn't test the non-ownership checking mode of the instrumentation.

### DIFF
--- a/backend/cn/bin/main.ml
+++ b/backend/cn/bin/main.ml
@@ -857,7 +857,8 @@ module Testing_flags = struct
 
   let gen_backtrack_attempts =
     let doc =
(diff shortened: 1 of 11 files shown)


ZippeyKeys12:
@rbanerjee20 You might find this interesting. Running [`abs.fail.c`](https://github.com/rems-project/cerberus/blob/731c4d23661f16fab423daf8f88ed69f0b3a0205/tests/cn-test-gen/src/abs.fail.c) from #441 fails for me locally, but not on [CI](https://github.com/rems-project/cerberus/actions/runs/11292949859/job/31410029318). I removed the test from this PR, but I don't know what to make of it.

PR #635, opened 2024-10-18
[CN-Exec/CN-Test-Gen] Generate functions for assuming ownership

ZippeyKeys12:
Allows separating generation and assuming ownership.
Would make memory management much simpler and cleaner for test case generation.

### DIFF
--- a/backend/cn/bin/main.ml
+++ b/backend/cn/bin/main.ml
@@ -496,7 +496,13 @@ let run_tests
               interactive
             }
           in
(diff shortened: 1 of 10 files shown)


ZippeyKeys12:
@rbanerjee20 I was originally going to put these functions in `cn.c`, but it causes issues interacting with globals. As a result, this is all in test generation, the only change to the instrumentation is renaming functions `check_owned_`, instead of `owned_` (since there are now `assume_owned_`).

PR #650, opened 2024-10-24
Ubi9 based Dockerfile

podhrmic:
This is a draft PR, exploring how to build CN on top of [Red Hat Ubi9](https://www.redhat.com/en/blog/introducing-red-hat-universal-base-image). I needed to use [Nix](https://nixos.org/) package manager to install Opam, since Opam does not have binary for Redhat/Fedora. This will need to be revisited in the future.

The build does error on `make install_cn` step with:
```
File "backend/cn/lib/testGeneration/genOptimize.ml", line 289, characters 30-49:
289 |                | Unsigned -> (Z.to_int64_unsigned, Z.of_int64_unsigned)
                                    ^^^^^^^^^^^^^^^^^^^
Error: Unbound value "Z.to_int64_unsigned"
make: *** [Makefile:79: cn] Error 1
```

I suspect this is because the installed Opam is too old. Any help here is appreciated.

### DIFF
--- a/Dockerfile.ubi
+++ b/Dockerfile.ubi
@@ -4,7 +4,7 @@ FROM redhat/ubi9:latest
 RUN yum update -y && \
     yum install -y xz sudo gcc unzip \
     diffutils patch pkgconfig bzip2 \
-    git perl wget
+    git perl wget ca-certificates
 
 # Install additional FEDORA packages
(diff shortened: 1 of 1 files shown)


podhrmic:
Closing in favor of #660

PR #656, opened 2024-10-25
Build multiplatform docker image

peterohanley:
The current build builds two single-platform images and then overwrites one with the other by pushing both to the same name.


peterohanley:
Looks like it won't run CI, and there's #660

PR #660, opened 2024-10-25
Build a truly multi platform docker image

### LINKED ISSUE #659: [BUG] the docker multiplatform image not correctly built
Right now, even though we are building arm64 image, we never push it. Add a proper image manifest.

podhrmic:
Fixes #659 
Follows the instructions from https://depot.dev/blog/multi-platform-docker-images-in-github-actions


podhrmic:
FYI I am not sure how to trigger the CI from a fork

Dockerfile.deps
```diff
@@ -1,18 +1,72 @@
...
+# add `--disable-sandboxing` which is needed to run as root
+RUN echo $COMPILER_VERSION
+RUN opam init --yes --no-setup --shell=sh --compiler=$COMPILER_VERSION --disable-sandboxing
+RUN opam switch create with_coq $COMPILER_VERSION && \
+  eval `opam env --switch=with_coq` && \
+  opam repo add --yes --this-switch coq-released https://coq.inria.fr/opam/released && \
+  opam pin --yes -n coq-struct-tact https://github.com/uwplse/StructTact.git && \
+  opam repo add --yes --this-switch iris-dev https://gitlab.mpi-sws.org/iris/opam.git && \
+  opam pin --yes -n coq-sail-stdpp https://github.com/rems-project/coq-sail.git#f319aad && \
+  opam pin --yes -n coq-cheri-capabilities https://github.com/rems-project/coq-cheri-capabilities.git
```
dc-mak: This is unnecessary, very slow and should be deleted.

Dockerfile.deps
```diff
@@ -1,18 +1,72 @@
...
+# such as a makefile target
 
+# Build matrix - compiler version
+ARG COMPILER_VERSION=4.14.1
+
+# System dependencies (ubuntu)
 RUN apt-get update
 RUN apt-get upgrade -y
-RUN apt-get install -y opam libgmp-dev libmpfr-dev
+RUN apt-get install -y build-essential libgmp-dev z3 opam cmake pkg-config python3-distutils
```
dc-mak: I believe pkg-config is required for zarith, and build-essential comes via opam, but the rest I'm not sure about?
Also, there's a [known-bug with the default Z3 version in Ubuntu 22.04](https://github.com/rems-project/cerberus/commit/efc7cdbd28bcb63f4d8b9b0da72f701a1e985f2d) so maybe a more recent binary release, or installing via opam, is more sensible (opam has version 4.13.0-3).

Dockerfile.deps
```diff
@@ -1,18 +1,72 @@
...
+RUN opam switch $COMPILER_VERSION && \
+  eval `opam env --switch=$COMPILER_VERSION` && \
+  opam pin --yes --no-action add cerberus-lib . && \
+  opam pin --yes --no-action add cerberus . && \
+  opam install --yes cerberus
+
+# Run Cerberus CI tests
+RUN opam switch $COMPILER_VERSION && \
+  eval `opam env --switch=$COMPILER_VERSION` && \
+  cd tests; USE_OPAM='' ./run-ci.sh
```
dc-mak: These are also unnecessary, you should be able to skip straight to the CN section - it will pick up the dependencies automatically.

Dockerfile.deps
```diff
@@ -1,18 +1,72 @@
...
+  chmod +x cvc5-Linux-static/bin/cvc5 && \
+  mv cp cvc5-Linux-static/bin/cvc5 /usr/local/bin/ && \
+  rm -rf cvc5-1.1.2.zip
+
+# Install CN
+WORKDIR /opt/cerberus
+RUN opam switch $COMPILER_VERSION && \
+  eval `opam env --switch=$COMPILER_VERSION` && \
+  opam pin --yes --no-action add cn . && \
+  opam install --yes cn ocamlformat.0.26.2
```
dc-mak: Is the Dockerfile being used for development or as an application? You won't need ocamlformat as an application, but you will for development (and the later source code formatting).

Dockerfile.deps
```diff
@@ -1,18 +1,72 @@
...
+
+# Check CN code formatting
+RUN opam switch $COMPILER_VERSION && \
+  eval `opam env --switch=$COMPILER_VERSION` && \
+  USE_OPAM='' cd backend/cn && dune build @fmt
+
+# Run CN CI tests
```
dc-mak: Ditto here.

dc-mak (commented):
What's the intended use case for this Dockerfile?

Dockerfile.deps
dc-mak: `opam install -y ./cerberus.opam /cerberus-lib ./cn.opam` should also work without pinning first.

podhrmic:
> What's the intended use case for this Dockerfile?

To run CN on OpenSUT code in a CI: https://github.com/GaloisInc/VERSE-OpenSUT (an example job is [here](https://github.com/GaloisInc/VERSE-OpenSUT/actions/runs/11620278295)). And more broadly to enable easy CN distribution, such that you don't need to install opam and build everything from scratch.

I think I can make the resulting image much smaller (just the compiled CN binary + basic OS dependencies should be sufficient).

dc-mak:
Ok in that case, it's more as an application than as dev environment. Simply doing `opam install -y ./cerberus.opam ./cerberus-lib.opam ./cn.opam` should suffice after system deps, opam and correct ocaml version have been installed.

That by itself should be fairly minimal. I don't think CN is linked statically, and so it will need its library dependencies.

podhrmic:
Correct, it is more an application in our use case. Do you think the steps outlined [here](https://ocamlpro.com/blog/2021_09_02_generating_static_and_portable_executables_with_ocaml/) would work for creating a statically linked binary? I am trying to avoid needing opam/ocaml for various reasons (see https://github.com/GaloisInc/VERSE-Toolchain/issues/48)

dc-mak:
It might work, but we have no experience with trying that and will be unable to help in case it doesn't. I don't see an explicit rejection of  the OCaml platform on the link you sent, and opam is availabe on RHEL. But obviously you can proceed as you need to; we can only support with build instructions as given.

podhrmic:
I tried running `opam install -y ./cerberus.opam ./cerberus-lib.opam ./cn.opam` but it resulted in *larger* images that when using the Makefile steps. 

I also added a RedHat image to the build.

podhrmic:
@dc-mak I believe the MR is now ready. The docker build works, but the image is not being pushed into the registry - I think this will be resolved for `master` builds. I also added more CI badges to the README.

Dockerfile.redhat
```diff
@@ -0,0 +1,40 @@
...
+ENV OPAMCONFIRMLEVEL=unsafe-yes
+RUN opam init --disable-sandboxing
+RUN opam install dune lem
+
+ADD . /opt/cerberus
+WORKDIR /opt/cerberus
+RUN opam install --deps-only  ./cerberus-lib.opam ./cn.opam
+
+RUN eval `opam env` \
+  && make install \
```
dc-mak: This line is not necessary.

Dockerfile.redhat
```diff
@@ -0,0 +1,40 @@
...
+    yum update -y && \
+    yum install -y z3
+
+# Install OPAM
+# See https://opam.ocaml.org/doc/1.2/Install.html
+RUN curl -fsSL https://opam.ocaml.org/install.sh | sh
+
+ENV OPAMCONFIRMLEVEL=unsafe-yes
+RUN opam init --disable-sandboxing
+RUN opam install dune lem
```
dc-mak: This line is unnecessary too.

PR #669, opened 2024-10-29
Fix #668 - pragma line loc info

### LINKED ISSUE #668: #pragma lines break source location information
For instance, given:
```c
#pragma clang diagnostic ignored "-Wunused-value"

int f () {
  assert(0);
}
```
`cn verify asd.c` produces

```
9:58 % cn verify asd.c
[1/1]: f -- fail
asd.c:3:3: error: Static error
int f () {
  ^~~~~~~~~
assert() failure
```
... where really it should be pointing at the `assert`.

dc-mak:


cp526:
That's great, thanks!

PR #670, opened 2024-10-29
CN VIP: Enable CI

dc-mak:


tests/cn_vip_testsuite/pointer_from_integer_1ig.annot.c
```diff
@@ -10,11 +10,10 @@ void f(uintptr_t i) {
 #if defined(ANNOT)
   int *p = copy_alloc_id(i, &j);
 #else
-  int *p = (int*)i;
+  int *p = (int*)i; // not even round-trip can save this
```
dc-mak, line 13: Comment should be clearer: "integer argument does not have provenance, so cannot roundtrip"

PR #675, opened 2024-10-30
[CN-Exec/CN-Test-Gen] Add allocation checkpoints

ZippeyKeys12:
With a bump allocator, we can store what its `curr` pointer is at each backtrack point. Then, since after backtracking, all values that were calculated or generated later are discarded, we can reset the `curr` pointer to what we had stored.

This means that generators don't leak memory due to backtracking.


ZippeyKeys12:
@rbanerjee20 this largely resolves memory issues for our input generators, where the only leaked memory is now from intermediate calculations that contribute to the actual result. Any intermediate values that don't (as-in they get forgotten during backtracking) are freed.

PR #679, opened 2024-10-31
Solver parameters for simple-smt interface + CVC5 logic parameter

cp526:


cp526:
This sets CVC5's logic to QF_AUFBVDTLIA, as recommended by Mathias Preiner.

cp526:
> This sets CVC5's logic to QF_AUFBVDTLIA, as recommended by Mathias Preiner.

Now actually QF_ALL, due to timeout on one test.

PR #681, opened 2024-10-31
CN: serialize JSON-encoded state traces on error

samcowger:
This introduces a new flag, `--json-trace`, to cause CN to serialize state trace files encoded as JSON, in addition to serializing them as HTML. It also exposes an API for libraries that depend on CN to parse these files.

### DIFF
--- a/backend/cn/lib/typeErrors.ml
+++ b/backend/cn/lib/typeErrors.ml
@@ -634,42 +634,47 @@ let mk_output_dir (output_dir : string option) : string =
       Filename.get_temp_dir_name ()
 
 
-(** Construct a canonical filename for state output derived from the given error
-    location, located in [output_dir], if possible. *)
-let mk_state_file_name
-  ?(output_dir : string option)
+(** A naming convention for files that pertain to specific error locations. The
+    generated name will always include the user-provided [~name], which should
+    be a valid filename. *)
+let located_file_name
   ?(fn_name : string option)
-  (loc : Cerb_location.t)
+  ~(dir : string)
+  ~(name : string)
+  ~(ext : string)
+  (error_loc : Cerb_location.t)
   : string
   =
-  let dir = mk_output_dir output_dir in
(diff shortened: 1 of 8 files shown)


backend/cn/lib/typeErrors.ml
```diff
@@ -650,35 +653,77 @@ let mk_state_file_name
   Filename.concat dir filename
 
 
+(** Construct a canonical filename for report output derived from the given
+    error location, located in [output_dir], if possible. *)
+let mk_report_file_name
```
dc-mak, line 671: Nit: might be more informative to rename this as `mk_json_file_name` (and the other one `mk_state_file_name => mk_html_file_name`? Actually you could even combine the two with a `type output_type = Html | Json` or `` [> `Html | `Json ]`` polymorphic datatype tag since they only seem to differ in their extensions?

cn.opam
```diff
@@ -7,6 +7,7 @@ depends: [
   "monomorphic"
   "ocaml" {>= "4.14.0"}
   "ppx_deriving"
+  "ppx_deriving_yojson"
```
dc-mak: Should the version you used be added as a lower bound?

dc-mak (commented):
Had a quick look, lgtm, few nits but happy to ship. Thanks so much for this work!

cn.opam
samcowger: Good call. It looks like CI picked 3.9.0, but it seems to have worked locally with 3.8.0, so I've specified the latter as a lower bound.

backend/cn/lib/typeErrors.ml
samcowger: I lifted out an abstraction along these lines - see 026ffbef2caa5ad86a505e1ff9f9dede0648c973. The filenames differ in extensions as well as "root" names ("state" for state files, "report" for report files), so I left the interface `string`-oriented for a little more flexibility.

PR #693, opened 2024-11-05
[CN-exec] Add a readme for Fulminate

rbanerjee20:
Include some basic instructions on how to use Fulminate, as per @yav 's request (and should be useful in general)


dc-mak:
Lgtm, could just be part of the README imo but it's not that big a deal either way.

PR #694, opened 2024-11-05
CN VIP: Add support for memcpy & proxy

dc-mak:
This commit hooks into an existing place for assigning function types to proxy functions. It assumes that memcpy's type is expressible in CN, which is true-ish for now, but may not remain so if we want to support round-trip for pointers without provenance in byte values, and thus need a "special type" or more arbitrary set of operations on the typing context to support.

### DIFF
--- a/backend/cn/lib/cLogicalFuns.ml
+++ b/backend/cn/lib/cLogicalFuns.ml
@@ -545,22 +545,23 @@ let rec symb_exec_expr ctxt state_vars expr =
       let loc, l_sym = SymMap.find nm ctxt.c_fun_pred_map in
       let@ def = get_logical_function_def loc l_sym in
       rcval (IT.apply_ l_sym args_its def.LogicalFunctions.return_bt loc) state)
-    else if Sym.has_id_with Setup.unfold_stdlib_name nm then (
-      let s = Option.get (Sym.has_id nm) in
-      let wrap_int x = IT.wrapI_ (signed_int_ity, x) in
-      if String.equal s "ctz_proxy" then
-        rcval
-          (wrap_int (IT.arith_unop Terms.BW_CTZ_NoSMT (List.hd args_its) loc) loc)
-          state
-      else if List.exists (String.equal s) [ "ffs_proxy"; "ffsl_proxy"; "ffsll_proxy" ]
-      then
-        rcval
-          (wrap_int (IT.arith_unop Terms.BW_FFS_NoSMT (List.hd args_its) loc) loc)
-          state
-      else
-        failwith ("unknown stdlib function: " ^ s))
-    else
-      fail_fun_it "not a function with a pure/logical interpretation"
+    else (
+      let bail = fail_fun_it "not a function with a pure/logical interpretation" in
+      match Sym.has_id nm with
+      | None -> bail
+      | Some s ->
+        let wrap_int x = IT.wrapI_ (signed_int_ity, x) in
+        if String.equal s "ctz_proxy" then
+          rcval
+            (wrap_int (IT.arith_unop Terms.BW_CTZ_NoSMT (List.hd args_its) loc) loc)
+            state
+        else if List.exists (String.equal s) [ "ffs_proxy"; "ffsl_proxy"; "ffsll_proxy" ]
+        then
+          rcval
+            (wrap_int (IT.arith_unop Terms.BW_FFS_NoSMT (List.hd args_its) loc) loc)
+            state
+        else
+          bail)
   | CN_progs _ -> rcval (IT.unit_ loc) state
   | _ ->
     fail_n
--- a/backend/cn/lib/check.ml
+++ b/backend/cn/lib/check.ml
@@ -1273,6 +1273,19 @@ let add_trace_information _labels annots =
   return ()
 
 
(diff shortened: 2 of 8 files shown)


backend/cn/lib/check.ml
```diff
@@ -1561,8 +1578,46 @@ let rec check_expr labels (e : BT.t Mu.expr) (k : IT.t -> unit m) : unit m =
...
+              let _, dest_arr = IT.fresh_named map_bt "dest_arr" here in
+              let@ () =
+                add_r
+                  loc
+                  ( Q (bytes_qpred (Sym.fresh_named "dest2") Sctypes.uchar_ct dest Init),
+                    O dest_arr )
+              in
+              let here = Locations.other __FUNCTION__ in
+              let@ () = add_c loc (LC.T (IT.eq_ (src_arr, dest_arr) here)) in
+              k (IT.unit_ here)))
```
dc-mak: Delete for now.

backend/cn/lib/sctypes.ml
```diff
@@ -134,6 +134,8 @@ let char_ct = Integer Char
 
 let uchar_ct = Integer (Unsigned Ichar)
 
+let size_t_ct = Integer Size_t
+
```
dc-mak: Delete

backend/cn/lib/typeErrors.ml
```diff
@@ -20,6 +20,8 @@ type access =
   | Free
   | To_bytes
   | From_bytes
+  | Memcpy
+  | Memcmp
```
dc-mak: Delete this too

backend/cn/lib/typeErrors.ml
```diff
@@ -67,6 +69,8 @@ let for_access = function
   | Free -> !^"for free-ing"
   | To_bytes -> !^"for converting to bytes"
   | From_bytes -> !^"for convering from bytes"
+  | Memcpy -> !^"for memcpy"
+  | Memcmp -> !^"for memcmp"
```
dc-mak: And this

tests/cn_vip_testsuite/cn_lemmas.h
```diff
@@ -36,17 +36,14 @@ void _memcpy(unsigned char *dest, unsigned char *src, size_t n);
...
-    (u64) dest <= (u64) dest + n;
     take Src = each (u64 i; 0u64 <= i && i < n ) { Owned(array_shift(src, i)) };
     take Dest = each (u64 i; 0u64 <= i && i < n ) { Block<unsigned char>(array_shift(dest, i)) };
 
 ensures
     take SrcR = each (u64 i; 0u64 <= i && i < n ) { Owned(array_shift(src, i)) };
     take DestR = each (u64 i; 0u64 <= i && i < n ) { Owned(array_shift(dest, i)) };
     Src == SrcR;
-    SrcR == DestR;
+    Src == DestR;
```
dc-mak: Delete this.

tests/cn_vip_testsuite/pointer_copy_memcpy.c
```diff
@@ -10,7 +10,7 @@ int main()
   int *q;
   /*CN_VIP*//*@ to_bytes Owned<int*>(&p); @*/
   /*CN_VIP*//*@ to_bytes Block<int*>(&q); @*/
-  _memcpy ((unsigned char*)&q, (unsigned char*)&p, sizeof p);
+  memcpy ((unsigned char*)&q, (unsigned char*)&p, sizeof p);
```
dc-mak: Fix spacing.

backend/cn/lib/cLogicalFuns.ml
```diff
@@ -545,7 +545,7 @@ let rec symb_exec_expr ctxt state_vars expr =
       let loc, l_sym = SymMap.find nm ctxt.c_fun_pred_map in
       let@ def = get_logical_function_def loc l_sym in
       rcval (IT.apply_ l_sym args_its def.LogicalFunctions.return_bt loc) state)
-    else if Sym.has_id_with Setup.unfold_stdlib_name nm then (
+    else (
```
dc-mak, line 548: There's still an option unwrapping here that needs to be guarded.

dc-mak: I think Cerberus elaboration guarantees that if you have a ccall with a symbol then it must be an `SD_id` ([and not the other cases of `Symbol.symbol_description`](https://github.com/rems-project/cerberus/blob/84d21f388def0fe34fc356a7741fabe1c6481156/frontend/model/symbol.lem#L77-L87))? But better to be defensive than crash anyways...
```
19:30 ➜  cerberus git:(cn-vip-memcpy) cat tmp.c
#include <string.h>

/*@ function (u32) foo(u32 x) @*/

unsigned int f(unsigned int x)
/*@ function foo; @*/
{
    memcpy(0, 0, 0);
}
19:30 ➜  cerberus git:(cn-vip-memcpy) cn verify tmp.c
tmp.c:8:5: error: getting expr from C syntax: function val: not a function with a pure/logical interpretation:
 a_712 : memcpy_proxy
    memcpy(0, 0, 0);
    ^~~~~~~~~~~~~~~
```

PR #699, opened 2024-11-06
[CN-Exec/CN-Test-Gen] Generalize failure handling

### LINKED ISSUE #697: "Out of memory!" while testing
Consider the following example:
```c
#include <stddef.h>

typedef int VEC_ELEMENT;

struct Vector {
  int size;
  int capacity;
  VEC_ELEMENT *data;
};


void *cn_malloc(size_t size);
void cn_free_sized(void *ptr, size_t size);

/*@

type_synonym VEC_ELEMENT = i32

type_synonym Vec = { i32 size, i32 capacity, map<i32,VEC_ELEMENT> elements }

predicate (Vec) Vec(pointer p) {
  take node = Owned<struct Vector>(p);
  assert(node.size <= node.capacity);
  take used = each(i32 i; 0i32 <= i && i < node.size) {
    Owned<VEC_ELEMENT>(array_shift<VEC_ELEMENT>(p,i))
  };
  take unused = each(i32 i; node.size <= i && i < node.capacity) {
    Block<VEC_ELEMENT>(array_shift<VEC_ELEMENT>(p,i))
  };
  return { size: node.size, capacity: node.capacity, elements: used };
} 
@*/


// Assumes malloc does not fail
struct Vector *vec_empty(int capacity)
/*@
requires
  0i32 <= capacity;
ensures
  take xs = Vec(return);
  xs.size == 0i32;
  xs.capacity == capacity;
@*/
{
  struct Vector *p = (struct Vector*) cn_malloc(sizeof(struct Vector));
  p->size = 0;
  p->capacity = capacity;
  p->data = (VEC_ELEMENT*) cn_malloc(sizeof(VEC_ELEMENT) * capacity);
  return p;
}
```

Trying to test this results in "out of memory".  This is not entirely surprising, as at the moment test generation only considers the precondition, and there's nothing there to limit the capacity of the empty vector.   This seems like a potentially common pattern for various "constructor" functions, so it might be good to have some plan for it.

One simple idea might be to have extra constraints used just for testing (e.g., and extra requires `capacity < 100` which is only used for testing but not proof)

ZippeyKeys12:
Generalizes failure handling to provide the reason for the failure. Also cleans up connect allocation to the failure mechanism.

For test-gen, this allows closes #697

### DIFF
--- a/runtime/libcn/include/cn-executable/utils.h
+++ b/runtime/libcn/include/cn-executable/utils.h
@@ -138,9 +138,18 @@ void cn_free_sized(void*, size_t len);
 void cn_print_nr_u64(int i, unsigned long u) ;
 void cn_print_u64(const char *str, unsigned long u) ;
 
-/* cn_exit callbacks */
-void set_cn_exit_cb(void (*callback)(void));
-void reset_cn_exit_cb(void);
+/* cn_failure callbacks */
+enum cn_failure_mode {
+    CN_FAILURE_ASSERT = 1,
+    CN_FAILURE_CHECK_OWNERSHIP,
+    CN_FAILURE_OWNERSHIP_LEAK,
+    CN_FAILURE_ALLOC
+};
+
+typedef void (*cn_failure_callback)(enum cn_failure_mode);
+void set_cn_failure_cb(cn_failure_callback callback);
+void reset_cn_failure_cb(void);
+void cn_failure(enum cn_failure_mode mode);
 
 /* Conversion functions */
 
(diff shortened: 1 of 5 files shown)


runtime/libcn/include/cn-executable/utils.h
```diff
@@ -138,9 +138,18 @@ void cn_free_sized(void*, size_t len);
...
+    CN_FAILURE_ASSERT = 1,
+    CN_FAILURE_CHECK_OWNERSHIP,
+    CN_FAILURE_OWNERSHIP_LEAK,
+    CN_FAILURE_ALLOC
+};
+
+typedef void (*cn_failure_callback)(enum cn_failure_reason);
+void set_cn_failure_cb(cn_failure_callback callback);
+void reset_cn_failure_cb(void);
+void cn_failure(enum cn_failure_reason reason);
```
rbanerjee20: maybe rename `reason` to `mode` (as in "failure modes")

rbanerjee20 (commented):
looks good, thanks for generalising this. have added one comment

PR #703, opened 2024-11-06
CN: Parallelise and caputure output in CN CI

### LINKED ISSUE #412: [CN] Capture all output (not just return codes) for CI in a convenient to update way

dc-mak:
Fixes #412


tests/run-cn.sh
dc-mak, line 1: Don't delete this just yet, https://github.com/rems-project/cerberus/pull/712 relies on it.

dc-mak:
The return codes aren't working properly

dc-mak:
The find command in the yml is messing with return code

PR #712, opened 2024-11-11
Fix z3 install and CN tests for Dockerfiles

podhrmic:
The nighly docker build hid an error in docker invocation. I fixed the invocation of `run-cn.sh` and properly installed z3 via opam in both docker images.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -2,13 +2,13 @@ name: docker
 
 on:
   # Run this action every day
-  schedule:
-    - cron: '30 18 * * *'
+  # schedule:
+  #   - cron: '30 18 * * *'
   # Run this action any time any dockerfile changes
   push:
     paths:
       - 'Dockerfile.**'
-      - '**/docker.yml'
+      - '**docker.yml'
 
 env:
   CERBERUS_IMAGE_ID: ghcr.io/rems-project/cerberus/cn
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -2,8 +2,8 @@ name: docker
 
 on:
   # Run this action every day
-  # schedule:
-  #   - cron: '30 18 * * *'
+  schedule:
+    - cron: '30 18 * * *'
   # Run this action any time any dockerfile changes
   push:
     paths:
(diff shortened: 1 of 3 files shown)


podhrmic:
@dc-mak this fixes the error in https://github.com/rems-project/cerberus/actions/runs/11784243588/job/32823035384

dc-mak:
Hmm, to minimise this in the future, can you please also adjust the docker.yml to run on any PR that edits the docker.yml or any Dockerfiles? https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request

podhrmic:
@dc-mak I tried - but as you can see, the on-change triggers do not seem to work very well (or at all). This matches my previous experience, so if you have an idea how to fix this go ahead, otherwise I would just merge as is, as moving forward we shouldn't need to change the Dockerfiles much

dc-mak:
It seems to have triggered the workflow but I did it when the actual scheduled job is also running, so it is waiting behind that to finish... https://github.com/rems-project/cerberus/actions/runs/11803628827

podhrmic:
@dc-mak I would also recommend that the `test-docker-images` job depends on the build job, such that you test the newly built images. 

Adding:
```
needs:
  needs: [docker-release-redhat, docker-release-ubuntu]
```

Indeed, this will only work on `master` branch, as you are limiting which branch can push the new images, but probably good enough for now

dc-mak:
> @dc-mak I would also recommend that the `test-docker-images` job depends on the build job, such that you test the newly built images.
> 
> Adding:
> 
> ```
> needs:
>   needs: [docker-release-redhat, docker-release-ubuntu]
> ```
> 
> Indeed, this will only work on `master` branch, as you are limiting which branch can push the new images, but probably good enough for now

Fair point, feel free to add and push

dc-mak:
https://docs.docker.com/build/ci/github-actions/share-image-jobs

https://docs.docker.com/reference/cli/docker/buildx/build/#output

There also seem to be some caching options of some sort.

podhrmic:
@dc-mak I updated the job dependency. I cannot use the caching, because I am building the images with attestation, and that is allowed only if you directly push the images into the registry (`attests: type=sbom` and `provenance: mode=max`). I think this is good enough for now, thanks for tweaking the push conditions!

podhrmic:
@dc-mak any reason not to merge this?

dc-mak:
I’ve been writing 2000 words/day.

PR #713, opened 2024-11-11
Improve and extend mechanism for extracting relevant information for executable checking

cp526:


rbanerjee20 (commented):
Looks good to me! Thanks for including the loop invariant info and also the loop condition location.

rbanerjee20:
Actually, seems there are some conflicts now, but probably after those are resolved it can be merged in

PR #754, opened 2024-12-10
Clarify SMT limitation

dc-mak:


backend/cn/lib/compile.ml
```diff
@@ -343,7 +342,9 @@ let add_datatype_info env (dt : cn_datatype) =
         { loc = Id.loc nm;
           msg =
             Generic
-              (!^"Re-using member name" ^^^ Id.pp nm ^^^ !^"within datatype definition.")
+              (!^"Re-using member name"
+               ^^^ Id.pp nm
+               ^^^ !^"within datatype definition (SMT limitation).")
```
cp526, line 347: That shouldn't be shown in the error message though, that's an implementation concern.

dc-mak: The idea was to discourage users from creating an issue about this/explain why CN has a somewhat strange and arbitrary restriction.

(I guess in theory a fancier implementation could deal with this, but in practice, just allowing it won't work.)

PR #756, opened 2024-12-11
CN VIP: Correct/explain shifting assumptions

dc-mak:
This is trickier than anticipated because of create_rdonly - it uses the effectful array shift for some reason.

### DIFF
--- a/backend/cn/lib/check.ml
+++ b/backend/cn/lib/check.ml
@@ -431,25 +431,14 @@ let check_has_alloc_id loc ptr ub_unspec =
     return ()
 
 
-let check_alloc_bounds loc ~ptr ub_unspec =
-  if !use_vip then (
-    let here = Locations.other __FUNCTION__ in
-    let module H = Alloc.History in
-    let H.{ base; size } = H.(split (lookup_ptr ptr here) here) in
-    let addr = addr_ ptr here in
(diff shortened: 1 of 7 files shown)


backend/cn/lib/check.ml
```diff
@@ -1509,50 +1482,45 @@ let rec check_expr labels (e : BT.t Mu.expr) (k : IT.t -> unit m) : unit m =
...
+          (* TODO (DCM, VIP): error if called on Void or Function Ctype *)
           check_pexpr pe (fun arg ->
-            let value = aligned_ (arg, act.ct) loc in
-            k value)
+            (* let unspec = CF.Undefined.UB_unspec_pointer_add in *)
+            (* let@ () = check_has_alloc_id loc arg unspec in *)
+            (* let index = num_lit_ Z.one Memory.uintptr_bt here in *)
+            (* let check_this = arrayShift_ ~base:arg ~index act.ct loc in *)
+            (* let ub = CF.Undefined.(UB_CERB004_unspecified unspec) in *)
+            (* let@ () = check_live_alloc_bounds `ISO_array_shift loc ub [ check_this ] in *)
```
dc-mak, line 1517: This is not quite right - the resource check may fail, but the type checking should succeed (and the value returned should be false).

PR #773, opened 2024-12-19
[CN-Exec] Extract and use CN-Test-Gen bounds analysis

### LINKED ISSUE #689: Failure while testing `each`
When I try to generate tests for the following code, I get an exception:
```c
void preserve(unsigned long size, int *p)                                           
/*@                                                                              
requires                                                                         
  take a1 = each(u64 i; i <= 0u64 && i < size) { Owned<int>(array_shift<int>(p,i)) };
ensures                                                                          
  take a2 = each(u64 i; i <= 0u64 && i < size) { Owned<int>(array_shift<int>(p,i)) };
@*/                                                                              
{                                                                                
} 
```
This throws:
```
 Failure("TODO rearrange_start_inequality at 6:25")
    Raised at Stdlib.failwith in file "stdlib.ml", line 29, characters 17-33
    Called from Cn__Cn_internal_to_ail.generate_start_expr in file "backend/cn/lib/cn_internal_to_ail.ml", line 533, characters 11-53
```
Another example that leads to the same exception is if you make the size an `int`, and add a cast:
```c
take a1 = each(u64 i; i <= 0u64 && i < (u64)size) { Owned<int>(array_shift<int>(p,i)) };
```

Another variation, if we remove the redundant `0 <= 0u64` constraint:
```c
take a1 = each(u64 i; i < size) { Owned<int>(array_shift<int>(p,i)) };
```
This throws:
```
cn: internal error, uncaught exception:
    Failure("Not of correct form (unlikely case - i's not matching)")
    Raised at Stdlib.failwith in file "stdlib.ml", line 29, characters 17-33
    Called from Cn__Cn_internal_to_ail.generate_start_expr in file "backend/cn/lib/cn_internal_to_ail.ml", line 530, characters 8-73
```

The only version I was able to run without exceptions is to use `int` as the size:
```c
take a1 = each(i32 i; 0i32 <= i && i < size) { Owned<int>(array_shift<int>(p,i)) };
```
This, however, leads to warnings from `cn` saying that `each` expects a `u64`.

ZippeyKeys12:
Extracts CN-Test-Gen's bounds analysis and uses it for CN-Exec, as previously discussed with Rini.

Covers many of the simplest cases, also adds nice error messages.

Closes #689.
There's always more generality possible, so new issues can be made as necessary for unsupported styles of `each` permissions.


ZippeyKeys12:
Need to update `tests/cn/to_from_bytes_block.c` to a successful test for `tests/run-cn-exec.sh`, since it is now supported.

PR #781, opened 2024-12-20
Fix #779: Mark offset off as an integer constant

### LINKED ISSUE #779: [CN] offsetof() not supported in integer constant expressions
```
typedef struct a {
    int b;
    int c;
} a;
_Static_assert(offsetof(a, c) == sizeof(int), "no gap");
```
```
cn verify  pre.c
pre.c:5:16: error: feature not yet supported: offsetof() in `integer constant expressions'
_Static_assert(offsetof(a, c) == sizeof(int), "no gap");
               ^~~~~~~~~~~~~~ 
```

I encountered these in [libxmss](https://github.com/FoxCryptoNL/xmss-library/blob/31df0b23556b119e91ca0193c77298cd71680ab4/src/signing_private.h#L689)

Is this a #406 issue, or is it doable but just not yet implemented? It would seem all the decisions had to be made to support `sizeof()` in ICE context.

dc-mak:


kmemarian:
@dc-mak can you also add the same match body to `is_arithmetic_constant_expression`?

kmemarian:
otherwise `offset()` still won't get accepted in initializers.

PR #782, opened 2024-12-21
increase maximum CN tuple arity, as needed by pgtable

cp526:


dc-mak:
@kmemarian can we please block merges with a failing CI? If we are concerned about wait times, we can use a merge queue 
https://github.com/orgs/community/discussions/75568#discussioncomment-7570257

dc-mak:
Or at the very least, enable auto-merge https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/automatically-merging-a-pull-request

kmemarian:
@dc-mak I think I've now added the status check for "build (4.14.1)". I doesn't offer the names of our workflows, but the categories they fall under. Does that work? I've also ticked the "Require branches to be up to date before merging".

Issue #678, opened 2024-10-31
[CN] cannot use prefix increment with size_t in a loop


peterohanley:
```
#include <stdlib.h>
#include <stdint.h>

void f(void)
{
    for (size_t i = 0; i < 10; i++);
    for (size_t i = 0; i < 10; ++i);
}
void g(void)
{
    for (uint64_t i = 0; i < 10; i++);
    for (uint64_t i = 0; i < 10; ++i);
}
```
```
% cn verify for_loop_u32.c
[1/2]: f -- fail
[2/2]: g -- pass
for_loop_u32.c:7:32: error: Mismatched types.
    for (size_t i = 0; i < 10; ++i);
                               ^~~ 
Expected value of type 'u64' but found value of type 'u32'
```

cerberus defines `size_t` as `__cerbty_size_t` and `uint64_t` as `unsigned long long` which I guess is where the discrepancy comes from.

peterohanley:
It looks like `__cerbty_size_t` just doesn't participate in integer promotion and it has nothing to do with loops:
```
% cn --version
git-c601ccf35-dirty [2024-12-30 16:32:05 -0500]
% cat pre.c
void f() {
    __cerbty_size_t a = 0;
    __cerbty_size_t b = 0+a;
}
% cn verify pre.c
[1/1]: f -- fail
pre.c:3:25: error: Mismatched types.
    __cerbty_size_t b = 0+a;
                        ~^~ 
Expected value of type 'u64' but found value of type 'u32'
```

Issue #272, opened 2024-05-27
[CN] Potentially incorrect typing `sizeof(uintptr_t) * CHAR_BIT`


dc-mak:
It's not clear why this fails type checking (note this is the same example as https://github.com/rems-project/cerberus/issues/270 but the issue here is not the bad location info, but why it fails in the first place).
Minimised and adapted from: https://github.com/rems-project/cerberus/blob/master/tests/pnvi_testsuite/pointer_copy_user_ctrlflow_bitwise.c
```
16:58 ➜  cerberus git:(master) cat illtyped_multiply.error.c && cn illtyped_multiply.error.c 2>tmp.out
#include<inttypes.h>
#include<limits.h>
#include<stddef.h>
size_t f(int x) {
    return sizeof(uintptr_t) * CHAR_BIT;
}
[1/1]: f
tests/cn/illtyped_multiply.error.c:5:12: error: Type error
    return sizeof(uintptr_t) * CHAR_BIT;
           ~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~
Expected value of type 'u64' but found value of type 'u32'
```

dc-mak:
Possibly related to https://github.com/rems-project/cerberus/issues/236 and also https://github.com/rems-project/cerberus/issues/257

dc-mak:
Similar issue happens for this 
```c
int
main()
{
        int size1 = sizeof(int);     // <- works
        size1 = size1 + 1;           // <- also works
        int size2 = sizeof(int) + 1; // <- fails
}
```

Taking a look at the Core, there is this line:
```ocaml
bound_op(wrap<size_t>, '+',  conv_int( ''unsigned int'',  a_640),  conv_int( ''unsigned int'',  a_641))
```
Specifically, CN infers `u32` for the `unsigned int` arguments, but expects `u64` for the `size_t` whole expression.

@kmemarian can you please clarify if this elaboration is as intended?  And while on the subject, I saw this comment in the CN code and I didn't understand it, or find any reference to it in your thesis - I think it's wrong? or irrelevant?
The code simply skips checking whether the type of pe1 equals the type of pe2 if the operator is a shift

https://github.com/rems-project/cerberus/blob/0bb93bf2adb99ba357f86b5d7ead1b55ca340e3a/backend/cn/lib/wellTyped.ml#L1767-L1781

dc-mak:
I think the following diff might be correct...

```diff
[...]

Issue #236, opened 2024-04-26
[CN] Type error on good C code.


scuellar:
**Description:**
CN reports a type error on correct C code. It seems like CN is making eager assumptions about the types of constants, which shouldn't be the case in C. 

**Steps to Reproduce:**
1. Create file `example.c` with
```
unsigned int type_error (unsigned int x) {
  return (-x)*0x4;
}
```
2. Run Cn
```
% cn example.c
```

**Expected Behavior:**
The file should type check correctly by `cn` and succeed. 

**Actual Behavior:**
The execution of `cn` fails, and the following error message is displayed:
```
error: Type error
  return (-x)*0x4;
         ^~~~ 
Expected value of type 'i32' but found value of type 'u32'
```

**Additional Information:**

- **`cn` Version**: unknown
- **Operating System:** OSx 13.5.1 (22G90)

septract:
I wonder if this example has the same root cause: 
```
int
main()
{
	if (sizeof(int) - 2 < 0)
		return 1;
	return 0;
}
```
(Mimized from [this example](https://github.com/rems-project/cn-tutorial/blob/main/src/example-archive/c-testsuite/broken/error-proof/00038.err1.c))

Result: 
```
broken/error-proof/00038.err1.c:8:6: error: Type error
        if (sizeof(int) - 2 < 0)
            ~~~~~~~~~~~~^~~ 
Expected value of type 'u64' but found value of type 'u32'
```

dc-mak:
Can you also add a regression test to the test directory for this?

cp526:
already there in cn-tutorial

dc-mak:
https://github.com/rems-project/cn-tutorial/commit/3453f6889c53af99f1bfad2f787d413907b8e483

PR #805, opened 2024-12-31
Adding a switch for normalising types at elaboration time

kmemarian:
This switch replaces type aliases (e.g. size_t, uintptr_t) with their implementation-defined choice during the elaboration to Core.

This commit enables the switch in the frontend of CN.

This should address some issues from #678, #236, #272.

@cp526 for #678 CN now gives a "Missing resource for reading" error on variables with the alias type.


kmemarian:
@cp526 the CI failures are the same

cp526:
Thanks very much @kmemarian , that looks great! 

The missing step to fix those CI failures is probably for CN to apply the same normalisation of C-types in certain places, so that where C-type equality is required (e.g. the resource inference), it uses equality post normalisation.

dc-mak:
Potentially also related to #815

PR #814, opened 2025-01-03
Add missing bits for supporting `default` in executable specs.

### LINKED ISSUE #714: [CN-Exec] Source injection error for single-line `if` statements with local variables
```c
void f()
{
    int found = 10;

    if (!found) return;
}

int main() {
    f();

    return 0;
}
```

Instrumentation of the `if` looks like:
```c
if (!CN_LOAD(found)) 
  c_remove_from_ghost_state((uintptr_t) &found, sizeof(signed int));
  goto __cn_epilogue;
```

It should be:

```c
if (!CN_LOAD(found)) {
  c_remove_from_ghost_state((uintptr_t) &found, sizeof(signed int));
  goto __cn_epilogue;
}
```

### LINKED ISSUE #813: Add support for `default` in executable specifications
Trying to use `default<u32>` throws `Failure("TODO Default")`.  We do have quite a lot of the infrastructure to support this,
we just need to finish it up.

yav:
Fixes #813.   The changes are:

* Replace the `todo` exception, with a call to the relevant `default` function.
* Remember to add the records from `default` to the record map.

Note: some future refactoring could be useful, but is not pressing, to factor out common stuff for calling `default` functions.

Fixes #714.  This also fixes an issue in #714 where we were missing some braces.   This should have been a separate PR but I accidentally pushed it here, and the change is sufficiently small that I felt it is not worth doing surgery on the repo to fix.

### DIFF
--- a/backend/cn/lib/cn_internal_to_ail.ml
+++ b/backend/cn/lib/cn_internal_to_ail.ml
@@ -576,7 +576,8 @@ let cn_to_ail_default bt =
   let do_call f = A.AilEcall (mk_expr (A.AilEident (Sym.fresh_pretty f)), []) in
   match get_underscored_typedef_string_from_bt bt with
   | Some f -> do_call ("default_" ^ f)
-  | None   -> failwith ("[UNSUPPORTED] default<" ^ Pp.plain (BT.pp bt) ^ ">")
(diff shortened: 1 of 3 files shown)


septract:
This works on the instances of `default` in the MKM proofs, nice! I should note that we only use `default` as a placeholder where we need an arbitrary value of the type, which is then thrown away. So I haven't tested whether `default` is correctly implemented, just that `cn test` can handle the construct at all.

PR #817, opened 2025-01-08
Add opam-versions because dune developer preview needs it

giltho:
I'm trying the [dune preview](https://preview.dune.build) for package management and it seems that it doesn't like empty dune files without an opam version... Feel free to ignore if you don't want that

See https://github.com/ocaml/dune/issues/11276


cp526:
Thanks!

PR #820, opened 2025-01-08
[CN-Exec] Add UBSan to CN-Exec CI

ZippeyKeys12:


ZippeyKeys12:
#825 adds UBSan w/o alignment (see #821)

Issue #345, opened 2024-06-24
[CN] Use a less arbitrary debug printing scheme


dc-mak:
E.g. https://erratique.ch/software/logs
```
12:58 ➜  cerberus git:(master) ✗ grep 'Pp.debug\s\+[0-9a-zA-Z]\+' -R backend/cn -o | awk -F':' '{ print $2; }' | sort | uniq -c
      6 Pp.debug 1
      1 Pp.debug 10
     15 Pp.debug 2
     10 Pp.debug 22
      8 Pp.debug 3
      3 Pp.debug 4
      7 Pp.debug 5
      2 Pp.debug 6
      3 Pp.debug 8
      9 Pp.debug 9
      1 Pp.debug i
      4 Pp.debug lvl
```

PR #829, opened 2025-01-14
CN: Check countermodels

cassiatorczon:
When the solver returns a countermodel, checks resources not given to the solver to see if their outputs in the countermodel could actually have come from those resources. (Undecidable in general, so this can return "unknown".)


bcpierce00:
I'm not an expert, but everything here looks OK to me!  Hopefully we can get it merged soon...

(I do see just a couple of TODO comments -- are those important, @cassiatorczon?)

backend/cn/lib/explain.ml
```diff
@@ -26,6 +26,202 @@ type log_entry =
...
+  (def : Def.Predicate.t)
+  (candidate : IT.t)
+  (ctxt : C.t)
+  (iarg_vals : IT.t list)
+  (term_vals : (IT.t * IT.t) list)
+  : LAT.check_result
+  =
+  (* ensure candidate type matches output type of predicate *)
+  if not (BT.equal (IT.get_bt candidate) def.oarg_bt) then
+    No
```
cp526: Given CN's type checking beforehand, will this ever happen?

backend/cn/lib/explain.ml
```diff
@@ -26,6 +26,202 @@ type log_entry =
...
+and check_clause
+  (c : Def.Clause.t)
+  (candidate : IT.t)
+  (ctxt : C.t)
+  (iargs : (Sym.t * BT.t) list)
+  (iarg_vals : IT.t list)
+  (term_vals : (IT.t * IT.t) list)
+  =
+  match Base.List.zip (List.map fst iargs) iarg_vals with
+  | Unequal_lengths -> Unknown !^"Wrong number of predicate arguments provided"
```
cp526: same: can this ever happen? If not, maybe this could be an assertion

backend/cn/lib/explain.ml
```diff
@@ -26,6 +26,202 @@ type log_entry =
...
+  in
+  (* use candidate to get terms for FVs in exp *)
+  match LAT.get_var_cands exp candidate with
+  | Yes var_cands -> f var_cands
+  | No e -> No e
+  | Error e -> Error e
+  | Unknown e ->
+    let s = Solver.make ctxt.global in
+    let loc = Cerb_location.unknown in
+    let _ = Solver.try_hard := true in
```
cp526: Do we have to reset Solver.try_hard after this?

backend/cn/lib/explain.ml
```diff
@@ -26,6 +26,202 @@ type log_entry =
...
+    (match line with
+     (* recurse with x's definition *)
+     | DefineL ((_, t), _) ->
+       get_body_constraints t var_def_locs v_cand ctxt iargs term_vals
+     | ResourceL ((_, (p, _)), _) ->
+       (match p with
+        | P psig ->
+          (match psig.name with
+           | Owned (_, _) ->
+             (* if the predicate is Owned, its pointer argument should not be null *)
```
cp526: How does this relate to the existing https://github.com/rems-project/cerberus/blob/74753593b25384d3079dcb0df7c58f4f8319bd08/backend/cn/lib/resource.ml#L25 and https://github.com/rems-project/cerberus/blob/74753593b25384d3079dcb0df7c58f4f8319bd08/backend/cn/lib/resource.ml#L52 ?

backend/cn/lib/logicalArgumentTypes.ml
```diff
@@ -227,3 +227,255 @@ let dtree dtree_i =
...
+    Yes (merge_eq eq acc combined)
+  in
+  match zip exps exps' with
+  | Some zipped -> List.fold_left merge_r_maps (Yes Sym.Map.empty) zipped
+  | None -> Error !^"Could not zip lists of expressions." (* should never happen *)
+
+
+(* Match an expression with free variables against a candidate returned by the solver to
+   get candidates for each of those free variables *)
+(* TODO: this is very naive right now;
```
cp526: Unless I'm misunderstanding – the way this is used, which way around will this be imprecise? I.e. will it sometimes say "Counterexample possible" when it's not, or "Counterexample impossible" when it is?

backend/cn/lib/request.ml
```diff
@@ -210,3 +210,7 @@ let alpha_equivalent r1 r2 =
 let steps_constant = function Q qp -> Option.is_some (IT.is_const qp.step) | _ -> true
 
 let dtree = function P pred -> Predicate.dtree pred | Q qpred -> QPredicate.dtree qpred
+
+let get_iargs = function P pred -> pred.iargs | Q qpred -> qpred.iargs
```
cp526: Depending on how this is used, this may be a bit dangerous: the quantifier in the `Q` case is in scope for `qpred.iargs`.

backend/cn/lib/solver.ml
```diff
@@ -1394,53 +1394,140 @@ let model_evaluator, reset_model_evaluator_state =
...
+
+
+  let translate_foralls solver assumptions =
+    List.map (translate_lc solver) (List.filter LC.is_forall assumptions)
+end
+
+let try_hard = ref false
+
 (** The main way to query the solver. *)
 let provable ~loc ~solver ~global ~assumptions ~simp_ctxt lc =
```
cp526: Are those the changes from the other pull request I made for the 'try hard' machinery? (Then I wouldn't have to check those again.)

backend/cn/lib/solver.ml
```diff
@@ -1394,53 +1394,140 @@ let model_evaluator, reset_model_evaluator_state =
...
+  (* assumes all constraints are bools*)
+  let not_forall acc lc = match lc with LC.T it -> it :: acc | _ -> acc in
+  let no_foralls = List.fold_left not_forall [] constraints in
+  let it_true = IT (Const (Bool true), BT.Bool, Cerb_location.unknown) in
+  let it_and acc lc = IT (Binop (And, acc, lc), BT.Bool, Cerb_location.unknown) in
+  List.fold_left it_and it_true no_foralls
+
+
+let ask_solver (s : solver) (lcs : LC.t list) : Simple_smt.result =
+  let smt_term = translate_term s (and_bool_constraints lcs) in
```
cp526: It's possible, but I don't know for sure, that mapping `translate_term` over all constraints and adding `assume`-ing them separately, as a flat list (rather than as a single nested `And`) would be better.

backend/cn/lib/logicalArgumentTypes.ml
```diff
@@ -227,3 +227,255 @@ let dtree dtree_i =
     | I i -> Dnode (pp_ctor "I", [ dtree_i i ])
   in
   aux
+
+
+(** Infrastructure for checking if a countermodel satisfies a predicate **)
```
cp526: Perhaps it'd be cleaner to move this into `Explain`?

bcpierce00:
Also, the CI seems to be failing for this branch at the moment.
image

ZippeyKeys12:
> Also, the CI seems to be failing for this branch at the moment. image

The second step of https://github.com/rems-project/cerberus/pull/865#issuecomment-2655199797 hadn't been done, should be fine now

backend/cn/lib/explain.ml
cassiatorczon: It shouldn't, but it seemed worth having as a sanity check. I'll change it to an assert, though.

backend/cn/lib/explain.ml
cassiatorczon: I'm not entirely sure how the solver machinery works, but I'll add that

backend/cn/lib/logicalArgumentTypes.ml
cassiatorczon: This specific function will return "unknown" on some inputs, so it will not be wrong either way. After this check, I run the solver to do a less useful version of the same check (it does not get constraints on the free variables), but that may say a counterexample is possible when it is not. For that reason and to avoid confusion, the display currently only shows resources that are definitely not valid. There may be some invalid ones not shown in the display.
I removed the "TODO" part of the comment, hopefully that's clearer.

backend/cn/lib/solver.ml
cassiatorczon: Yes, they are.

cassiatorczon:
> I'm not an expert, but everything here looks OK to me! Hopefully we can get it merged soon...
> 
> (I do see just a couple of TODO comments -- are those important, @cassiatorczon?)

The only one that may be important is the difference between `ctxt.logical` and `ctxt.computational` in explain.ml - @cp526 could you clarify the difference?

backend/cn/lib/explain.ml
cassiatorczon: Maybe I'm unfamiliar with OCaml best practices in this scenario -- `Base.List.Zip` returns a value of type `Base.List.Or_unequal_lengths.t`, which can be either `Unequal_lengths` or `Ok` with the zipped list, so I was matching to get the zipped list.

backend/cn/lib/explain.ml
cassiatorczon: It's possible this is duplicated -- the purpose of this case is to get any constraints we know about `p` if it was passed into `Owned`, so in particular, that it is not null. In the case where `p` was passed into a predicate, we recursively call `check_pred`. Could you clarify what `derived_lc1` and `derived_lc2` are doing?

backend/cn/lib/request.ml
cassiatorczon: quantified predicates are actually not in scope for my work, so I only use it for non-quantified predicates, but I can replace it with a match to avoid letting it be used elsewhere.

backend/cn/lib/solver.ml
cassiatorczon: Ok, I can change that. For future reference, why is that?

backend/cn/lib/explain.ml
cp526: `derived_lc1` and `derived_lc2` derive logical constraints from resource ownership; `derived_lc1` takes a single resource and derives (a) pointer range constraints and (b) `IT.hasAllocId_ pointer` which (I believe, @dc-mak ?) implies the pointer is not null; `derived_lc2` derives constraints implied by the simultaneous ownership of the two resources that it takes as arguments – resource "footprint" disjointness.

backend/cn/lib/explain.ml
cp526: Thanks! The `try_hard` does not persist across CN invocations, but it wouldn't get reset within a CN run. If your code only runs on a path leading to a "fatal" CN error, it might not matter right now, though in the future we may want CN to be able to "recover" from various kinds of errors rather than stop immediately.

backend/cn/lib/logicalArgumentTypes.ml
cp526: > For that reason and to avoid confusion, the display currently only shows resources that are definitely not valid.

That's great, that's exactly what we want, I think

cp526:
> > I'm not an expert, but everything here looks OK to me! Hopefully we can get it merged soon...
> > (I do see just a couple of TODO comments -- are those important, @cassiatorczon?)
> 
> The only one that may be important is the difference between `ctxt.logical` and `ctxt.computational` in explain.ml - @cp526 could you clarify the difference?

`ctxt.computational` contains variables that are present in the C program (or the Core program generated for it) – so variables that in some form are part of the computation; this is as opposed to `ctxt.logical`, which contains purely logical variables, introduced either by `take`'s in specifications, or internally by the type system's inference.

For your purposes, I expect, they can be treated just the same.

backend/cn/lib/solver.ml
cp526: No idea :-) But I seem to recall a small experiment where this might have been a performance difference.

backend/cn/lib/solver.ml
```diff
@@ -1394,53 +1394,138 @@ let model_evaluator, reset_model_evaluator_state =
...
+  s
+
+
+let ask_solver (s : solver) (lcs : LC.t list) : Simple_smt.result =
+  let s' = assume_constraints lcs s in
+  debug_ack_command s (SMT.push 1);
+  let simp_solver = s'.smt_solver in
+  let res = SMT.check simp_solver in
+  debug_ack_command s' (SMT.pop 1);
+  res
```
cp526: Do you ever re-use the solver `s` after running `ask_solver` once? 

The `push` and `pop` are not bracketed around the `assume_constraints`, so they won't get dropped after the `ask_solver` invocation, which may or may not be a problem.

cp526:
Great, modulo the question about `ask_solver` and your question about `ctxt.computational` vs `ctxt.logical` I think that's good to go.

backend/cn/lib/explain.ml
dc-mak: IIRC @cp526 did not want to add Base as a dependency (based on in-person conversation). (FWIW, I did want to: #347).

backend/cn/bin/main.ml
```diff
@@ -872,6 +874,11 @@ module Verify_flags = struct
       & info [ "solver-type" ] ~docv:"z3|cvc5" ~doc)
 
 
+  let try_hard =
+    let doc = "Try undecidable SMT solving using full set of assumptions" in
+    Arg.(value & flag & info [ "try_hard" ] ~doc)
```
dc-mak: Please use `try-hard` instead of `try_hard` (use dash instead of underscore).

dc-mak:
Apologies, please hold fire. I see issues and I'm reviewing right now.

backend/cn/lib/definition.ml
```diff
@@ -115,6 +115,25 @@ module Clause = struct
...
+      in
+      let so_far_ng = IT.and_ [ IT.not_ cur_guard cerb_loc; prev_negated ] cerb_loc in
+      cur :: explicit_negative_guards_aux cs' so_far_ng
+
+
+  let explicit_negative_guards (cs : t list) : t list =
+    let cerb_loc = Cerb_location.unknown in
+    match cs with
+    | [] -> []
+    | h :: tl -> h :: explicit_negative_guards_aux tl (IT.not_ h.guard cerb_loc)
```
dc-mak: * Use `let here = Locations.other __LOC__ in` instead of `Cerb_location.unknown`. This makes it much easier to debug.
* Is this purely for display, or is passed to the SMT solver? 
* Pretty sure this can be re-written with a `List.fold_right`?

backend/cn/lib/explain.ml
```diff
@@ -26,6 +26,430 @@ type log_entry =
...
+  type check_result = (LC.t list, Pp.document) result_with_data
+
+  (* let pp_check_result =
+     pp_result_with_data (Pp.list (fun lc -> !^"\n" ^^^ LC.pp lc)) (fun d -> d) *)
+
+  let filter_map_some (f : 'a -> 'b option) (l : 'a list) : 'b list =
+    List.fold_left
+      (fun acc elem -> match f elem with None -> acc | Some x -> x :: acc)
+      []
+      l
```
dc-mak: I'm confused, in what way is this not just [List.filter_map](https://ocaml.org/manual/5.2/api/List.html#VALfilter_map)?

backend/cn/lib/explain.ml
```diff
@@ -26,6 +26,430 @@ type log_entry =
 
 type log = log_entry list (* most recent first *)
 
+(** Infrastructure for checking if a countermodel satisfies a predicate **)
+module CheckPreds = struct
+  open ResultWithData
```
dc-mak: Avoid opening modules, use aliases and local opens if need be.

backend/cn/lib/explain.ml
```diff
@@ -26,6 +26,430 @@ type log_entry =
...
+    | DefineL of (Sym.t * IT.t) * Loc.info
+    | ResourceL of (Sym.t * (Req.t * BT.t)) * Loc.info
+
+  (* Optionally zip two lists, returning None if the lists have different lengths *)
+  let rec zip (l1 : 'a list) (l2 : 'b list) : ('a * 'b) list option =
+    match (l1, l2) with
+    | [], [] -> Some []
+    | h1 :: tl1, h2 :: tl2 ->
+      (match zip tl1 tl2 with Some zs -> Some ((h1, h2) :: zs) | None -> None)
+    | _, _ -> None
```
dc-mak: Please place in `backend/cn/list.ml`? Also you can use `List.map2` and just catch the `Invalid_argument` to return `None`.

backend/cn/lib/explain.ml
```diff
@@ -26,6 +26,430 @@ type log_entry =
...
+        match (acc, res) with
+        | Yes l, _ -> Yes l
+        | _, Yes l -> Yes l
+        | Error s, _ -> Error s
+        | _, Error s -> Error s
+        | Unknown s, _ -> Unknown s
+        | _, Unknown s -> Unknown s
+        | No s, _ -> No s
+      in
+      List.fold_left combine h t
```
dc-mak: Might this be better placed in `resultWithData.ml`?

backend/cn/lib/explain.ml
```diff
@@ -26,6 +26,430 @@ type log_entry =
...
+    | _, _ -> None
+
+
+  (* Take the union of two symbol maps,
+     removing any key that is in both maps but has a different value in each *)
+  let merge_eq (eq : 'a -> 'a -> bool) (m1 : 'a Sym.Map.t) (m2 : 'a Sym.Map.t)
+    : 'a Sym.Map.t
+    =
+    let merge _ v1 v2 = if eq v1 v2 then Some v1 else None in
+    Sym.Map.union merge m1 m2
```
dc-mak: Type annotations are fine for development and complicated functions, IMO for simple functions like this they add clutter.

backend/cn/lib/explain.ml
```diff
@@ -26,6 +26,430 @@ type log_entry =
...
+  (* ask the solver if the given set of constraints is satisfiable *)
+  let ask_solver g lcs =
+    match Solver.ask_solver (Solver.make g) lcs with
+    | Unsat -> No !^"Solver returned No."
+    | Unknown -> Unknown !^"Solver returned Unknown."
+    | Sat -> Yes lcs
+
+
+  let pair_to_lc (ps : IT.t * IT.t) : LogicalConstraints.t =
+    let loc = Cerb_location.unknown in
```
dc-mak: `let here = Locations.other __LOC__`

backend/cn/lib/explain.ml
```diff
@@ -26,6 +26,430 @@ type log_entry =
...
+    LC.T (IT.eq_ (fst ps, snd ps) loc)
+
+
+  (* convert a list of variable assignments to equality constraints *)
+  let convert_symmap_to_lcs (m : IT.t Sym.Map.t) : LogicalConstraints.t list =
+    let loc = Cerb_location.unknown in
```
dc-mak: `let here = Locations.other __LOC__`

backend/cn/lib/explain.ml
```diff
@@ -26,6 +26,430 @@ type log_entry =
...
+      List.fold_left accumulate_results (Yes ([], var_cands)) (Sym.Map.bindings var_cands)
+    in
+    (* use candidate to get terms for FVs in exp *)
+    match get_var_cands exp candidate with
+    | Yes var_cands -> f var_cands
+    | No e -> No e
+    | Error e -> Error e
+    | Unknown e ->
+      let s = Solver.make ctxt.global in
+      let loc = Cerb_location.unknown in
```
dc-mak: `let here = Locations.other __LOC__`

backend/cn/lib/explain.ml
```diff
@@ -26,6 +26,430 @@ type log_entry =
...
+  and get_var_constraints
+    (v : Sym.t)
+    (v_cand : IT.t)
+    (var_cands : IT.t Sym.Map.t)
+    (var_def_locs : def_line Sym.Map.t)
+    (ctxt : C.t)
+    (iargs : (Sym.t * BT.t) list)
+    (term_vals : (IT.t * IT.t) list)
+    =
+    let loc = Cerb_location.unknown in
```
dc-mak: `let here = Locations.other __LOC__`

backend/cn/lib/explain.ml
```diff
@@ -317,7 +748,46 @@ let state (ctxt : C.t) log model_with_q extras =
...
+    let check (rt, o) =
+      match (rt, o) with
+      | Req.Q _, _ -> None
+      | Req.P { name = Owned _; pointer = _; iargs = _ }, _ -> None
+      | Req.P { name = PName s; pointer = _; iargs }, Resource.O it ->
+        (match (Sym.Map.find_opt s defs, evaluate it) with
+         | Some def, Some cand ->
+           let ptr_val = Req.get_pointer rt in
+           let ptr_def =
+             (IT.sym_ (def.pointer, IT.get_bt ptr_val, Cerb_location.unknown), ptr_val)
```
dc-mak: `let here = Locations.other __LOC__`

backend/cn/lib/report.ml
```diff
@@ -36,6 +36,12 @@ let lab_interesting : label = "interesting"
 
 let lab_uninteresting : label = "uninteresting"
 
+let lab_invalid : label = "Resources that do not satisfy predicate definitions"
+
+let lab_unknown : label = "Resources that may not satisfy predicate definitions"
+
+let lab_valid : label = "Resources that do satisfy predicate definitions"
+
```
dc-mak: No need for type annotations. Also consider a datatype instead of using strings, and then converting to string if these labels are going to be used frequently (to avoid stringly-typed error messages/reports, to make it easier to add different error reporting backends).

backend/cn/lib/report.ml
```diff
@@ -232,6 +239,29 @@ let simp_view s =
...
+let make_invalid_resources rs =
+  h
+    1
+    lab_invalid
+    (table_by_label
+       table_without_head
+       simp_view
+       rs
+       lab_invalid
+       [] (* [ lab_unknown; lab_valid ] *))
```
dc-mak: If the comment needs to stick around, then please explain more, otherwise delete.

backend/cn/lib/solver.ml
```diff
@@ -1394,53 +1394,138 @@ let model_evaluator, reset_model_evaluator_state =
...
+module TryHard = struct
+  let translate_forall solver qs body =
+    let rec alpha_rename qs body =
+      match qs with
+      | [] -> ([], body)
+      | (s, bt) :: qs' ->
+        let qs', body = alpha_rename qs' body in
+        let s, body = IT.alpha_rename s body in
+        ((s, bt) :: qs', body)
+    in
```
dc-mak: Fairly sure this is a `List.fold_right` too.

backend/cn/lib/solver.ml
```diff
@@ -1394,53 +1394,138 @@ let model_evaluator, reset_model_evaluator_state =
...
+
 (** The main way to query the solver. *)
 let provable ~loc ~solver ~global ~assumptions ~simp_ctxt lc =
   let _ = loc in
-  (* ISD: should we use this somehow? *)
-  let s1 = { solver with globals = global } in
-  let rtrue () =
-    model_state := No_model;
-    `True
+  let _ = global in
```
dc-mak: If this is because the `~global` is unused, either (a, preferred) delete it or (b, if deleting is too complicated) do `~global:_`

backend/cn/lib/solver.ml
```diff
@@ -1394,53 +1394,138 @@ let model_evaluator, reset_model_evaluator_state =
...
+
+  let translate_foralls solver assumptions =
+    List.map (translate_lc solver) (List.filter LC.is_forall assumptions)
+end
+
+let try_hard = ref false
+
 (** The main way to query the solver. *)
 let provable ~loc ~solver ~global ~assumptions ~simp_ctxt lc =
   let _ = loc in
```
dc-mak, line 1452: See `let _ = global`. I know you didn't write it, but doing small clean-up when you're in the neighbourhood adds up.

backend/cn/lib/solver.ml
```diff
@@ -1394,53 +1394,138 @@ let model_evaluator, reset_model_evaluator_state =
...
+    debug_ack_command solver (SMT.push 1);
+    debug_ack_command solver (SMT.assume (SMT.bool_ands (nexpr :: extra)));
+    (match SMT.check inc with
      | SMT.Unsat ->
-       debug_ack_command s1 (SMT.pop 1);
-       rtrue ()
+       debug_ack_command solver (SMT.pop 1);
+       model_state := No_model;
+       `True
+     | SMT.Sat when !try_hard ->
```
dc-mak, line 1473: Would you be willing to use an if-expression? Guards like this disable exhaustive pattern-matching checks (even though admittedly, in this case, the stakes are smaller since the type is unlikely to change).

If you can lift the try-hard code into a separate function that will also help.

backend/cn/lib/solver.ml
```diff
@@ -1394,53 +1394,138 @@ let model_evaluator, reset_model_evaluator_state =
...
+          `True
+        | SMT.Sat ->
+          set_model inc qs;
+          debug_ack_command solver (SMT.pop 1);
+          Pp.(debug 3 (lazy !^"***** try-hard: unprovable *****"));
+          `False
+        | SMT.Unknown ->
+          set_model inc qs;
+          debug_ack_command solver (SMT.pop 1);
+          Pp.(debug 3 (lazy !^"***** try-hard: unknown *****"));
```
dc-mak: If the debug code is no-longer needed, please can you delete it? Our debug output is already quite cluttered (#345)

backend/cn/lib/solver.ml
```diff
@@ -1394,53 +1394,138 @@ let model_evaluator, reset_model_evaluator_state =
...
-   reason) *)
-
 (* ISD: Could these globs be different from the saved ones? *)
 let eval mo t =
   let model_fn = Hashtbl.find models_tbl mo in
   model_fn t
+
+
+let assume_constraints (constraints : LC.t list) (s : solver) : solver =
+  let get_term lc = match lc with LC.T it -> Some (translate_term s it) | _ -> None in
```
dc-mak: Nit: avoid wildcards in pattern-matches unless good reason.

backend/cn/lib/solver.ml
```diff
@@ -1394,53 +1394,138 @@ let model_evaluator, reset_model_evaluator_state =
...
+  let terms = List.filter_map get_term constraints in
+  let add_term term = debug_ack_command s (SMT.assume term) in
+  let _ = List.iter add_term terms in
```
dc-mak: Use a semicolon

backend/cn/lib/solver.mli
```diff
@@ -56,3 +56,8 @@ val model : unit -> model_with_q
     counter-example model (e.g. for evaluating sub-terms). Might return None in
     case we ask for the value of a "don't care" value in the (minimal) model. *)
 val eval : model -> IndexTerms.t -> IndexTerms.t option
+
+val ask_solver : solver -> LogicalConstraints.t list -> Simple_smt.result
```
dc-mak: Can you add a doc comment for this? How is it different to `provable`? Can you generalise provable and use that instead?

backend/cn/lib/solver.ml
```diff
@@ -1394,53 +1394,138 @@ let model_evaluator, reset_model_evaluator_state =
...
-    debug_ack_command s1 (SMT.push 1);
-    debug_ack_command s1 (SMT.assume (SMT.bool_ands (nlc :: extra)));
-    let res = SMT.check inc in
-    (match res with
+    let { expr; qs; extra } = translate_goal solver assumptions lc in
+    let nexpr = SMT.bool_not expr in
+    let inc = solver.smt_solver in
+    debug_ack_command solver (SMT.push 1);
+    debug_ack_command solver (SMT.assume (SMT.bool_ands (nexpr :: extra)));
+    (match SMT.check inc with
```
dc-mak, line 1468: For future ease of reviewing, it would be helpful if renaming and refactoring changes could be separated (into different PRs, or at least commits) from functional ones. These lines are are much harder to check. I appreciate it is more work to write, but it also makes digging through the commit history a lot easier, and that is read a lot more than it is written.

Happy to assist with git-fu if need be.

backend/cn/lib/explain.ml
```diff
@@ -317,7 +748,46 @@ let state (ctxt : C.t) log model_with_q extras =
...
+          (*original = ^^^ !^"\n"^^^ LAT.pp_check_result p; *)
+          simplified = [ rslt ]
+        }
+    in
+    Rp.add_labeled Rp.lab_invalid (List.map pp_checked_res nos) Rp.labeled_empty
+    (* Currently only displays invalid predicates *)
+    (* (Rp.add_labeled
+       Rp.lab_unknown
+       (List.map pp_checked_res unknown)
+       (Rp.add_labeled Rp.lab_valid (List.map pp_checked_res yeses) Rp.labeled_empty)) *)
```
dc-mak: Please delete if not necessary or explain if so.

backend/cn/lib/explain.ml
```diff
@@ -317,7 +748,46 @@ let state (ctxt : C.t) log model_with_q extras =
...
+    in
+    let checked = CheckPreds.filter_map_some check (C.get_rs ctxt) in
+    let nos, _ = List.partition (fun (r, _, _) -> ResultWithData.is_no r) checked in
+    (* let yeses, unknown = List.partition (fun (r, _, _) -> is_yes r) rest in *)
+    let pp_checked_res (p, req, cand) =
+      let _ = p in
+      let rslt = Req.pp req ^^^ !^"(" ^^^ IT.pp cand ^^^ !^")" in
+      Rp.
+        { original = rslt;
+          (*original = ^^^ !^"\n"^^^ LAT.pp_check_result p; *)
```
dc-mak: Please delete if not necessary or explain if so.

backend/cn/lib/explain.ml
```diff
@@ -317,7 +748,46 @@ let state (ctxt : C.t) log model_with_q extras =
...
+           in
+           Some (CheckPreds.check_pred s def cand ctxt iargs (ptr_def :: vals), rt, it)
+         | Some _, None ->
+           Some (Error (!^"Could not locate definition of variable" ^^^ IT.pp it), rt, it)
+         | None, _ ->
+           Some (Error (!^"Could not locate definition of predicate" ^^^ Sym.pp s), rt, it))
+    in
+    let checked = CheckPreds.filter_map_some check (C.get_rs ctxt) in
+    let nos, _ = List.partition (fun (r, _, _) -> ResultWithData.is_no r) checked in
+    (* let yeses, unknown = List.partition (fun (r, _, _) -> is_yes r) rest in *)
```
dc-mak, line 349: Please delete if not necessary or explain if so.

backend/cn/lib/explain.ml
```diff
@@ -26,6 +26,430 @@ type log_entry =
...
+    (* find def of x *)
+    match Sym.Map.find_opt v var_def_locs with
+    | None ->
+      (match (Sym.Map.find_opt v ctxt.logical, Sym.Map.find_opt v ctxt.computational) with
+       | Some (Value it, _), _ ->
+         get_body_constraints it var_def_locs v_cand ctxt iargs term_vals
+       | _, Some (Value it, _) ->
+         get_body_constraints it var_def_locs v_cand ctxt iargs term_vals
+       (* TODO: logical vs computational *)
+       (* TODO: BaseType case *)
```
dc-mak: Please make an issue for this rather than using TODOs.

backend/cn/lib/explain.ml
```diff
@@ -26,6 +26,430 @@ type log_entry =
...
+    (v : Sym.t)
+    (v_cand : IT.t)
+    (var_cands : IT.t Sym.Map.t)
+    (var_def_locs : def_line Sym.Map.t)
+    (ctxt : C.t)
+    (iargs : (Sym.t * BT.t) list)
+    (term_vals : (IT.t * IT.t) list)
+    =
+    let loc = Cerb_location.unknown in
+    (* find def of x *)
```
dc-mak: Delete?

backend/cn/lib/explain.ml
```diff
@@ -26,6 +26,430 @@ type log_entry =
...
+      let ln = ResourceL ((v, (rt, bt)), i) in
+      let new_defs = Sym.Map.add v ln defs in
+      organize_lines_aux next new_defs lcs
+    | Constraint (lc, _, next) -> organize_lines_aux next defs (lc :: lcs)
+    | I it -> (it, defs, lcs)
+
+
+  (* Sort lines into the returned expression, a map of variables to their defining lines, and a list of constraints *)
+  let organize_lines (lines : LAT.packing_ft) : IT.t * def_line Sym.Map.t * LC.t list =
+    organize_lines_aux lines Sym.Map.empty []
```
dc-mak: This is also probably a List.fold_right?

backend/cn/lib/resultWithData.ml
dc-mak, line 1: I think this can be replaced with ``('a, [`No | `Unknown | `Error] * 'b) Result.t``.

backend/cn/lib/explain.ml
```diff
@@ -26,6 +26,430 @@ type log_entry =
...
 type log = log_entry list (* most recent first *)
 
+(** Infrastructure for checking if a countermodel satisfies a predicate **)
+module CheckPreds = struct
+  open ResultWithData
+
+  type check_result = (LC.t list, Pp.document) result_with_data
+
+  (* let pp_check_result =
+     pp_result_with_data (Pp.list (fun lc -> !^"\n" ^^^ LC.pp lc)) (fun d -> d) *)
```
dc-mak: Please delete.

backend/cn/lib/explain.ml
```diff
@@ -26,6 +26,430 @@ type log_entry =
...
+    : 'a Sym.Map.t
+    =
+    let merge _ v1 v2 = if eq v1 v2 then Some v1 else None in
+    Sym.Map.union merge m1 m2
+
+
+  (* Build a map by using f to develop a map for each
+     pair of elements in the two lists, failing
+     if they produce different results for any symbol or if
+     the lists have different lengths *)
```
dc-mak: I'm struggling to parse this sentence.

backend/cn/lib/explain.ml
```diff
@@ -26,6 +26,430 @@ type log_entry =
...
+  let map_from_lists f eq exps exps' =
+    let merge_r_maps r_acc (exp1, exp1') =
+      let@ acc = r_acc in
+      let@ combined = f exp1 exp1' in
+      Yes (merge_eq eq acc combined)
+    in
+    match zip exps exps' with
+    | Some zipped -> List.fold_left merge_r_maps (Yes Sym.Map.empty) zipped
+    | None -> Error !^"Could not zip lists of expressions." (* should never happen *)
```
dc-mak: You can do `assert false` if you're sure about this.

backend/cn/lib/explain.ml
```diff
@@ -26,6 +26,430 @@ type log_entry =
...
+
+
+  (* Match an expression with free variables against a candidate returned by the solver to
+     get candidates for each of those free variables *)
```
dc-mak: I'm struggling to parse this sentence too. Is the word "candidate" is being used in two different ways here?

backend/cn/lib/explain.ml
```diff
@@ -26,6 +26,430 @@ type log_entry =
 
 type log = log_entry list (* most recent first *)
 
+(** Infrastructure for checking if a countermodel satisfies a predicate **)
```
dc-mak: Can you add a little bit more about the terminology and strategy here too? The key functions to look at, how they relate to each other.

backend/cn/lib/explain.ml
```diff
@@ -26,6 +26,430 @@ type log_entry =
...
+               Yes ([ LC.T neq ], var_cands)
+             | PName name ->
+               (* search for predicate definition *)
+               (match Sym.Map.find_opt name ctxt.global.resource_predicates with
+                | Some pdef ->
+                  (match check_pred name pdef v_cand ctxt psig.iargs term_vals with
+                   | Yes cs -> Yes (cs, var_cands)
+                   | No e -> No e
+                   | Unknown e -> Unknown e
+                   | Error e -> Error e)
```
dc-mak: Looks like a `check_pred name ... |> map (fun cs -> (cs, var_cands))` ?

backend/cn/lib/explain.ml
```diff
@@ -26,6 +26,430 @@ type log_entry =
...
+          | _ -> Unknown (!^"Could not find variable definition line for" ^^^ Sym.pp v)))
+    | Some line ->
+      (match line with
+       (* recurse with x's definition *)
+       | DefineL ((_, t), _) ->
+         get_body_constraints t var_def_locs v_cand ctxt iargs term_vals
+       | ResourceL ((_, (p, _)), _) ->
+         (match p with
+          | P psig ->
+            (match psig.name with
```
dc-mak: You can match directly on the record and field to reduce the nesting and make the code more readable:
```ocaml
| P { name = Owned _; (* remaining fields *) } ->
  (* code *)
| P { name = PName name; (* remaining fields *) } ->
  (* other code *)
| Q _ -> 
  (* more code *)
```

backend/cn/lib/explain.ml
```diff
@@ -26,6 +26,430 @@ type log_entry =
...
+       | _, Some (Value it, _) ->
+         get_body_constraints it var_def_locs v_cand ctxt iargs term_vals
+       (* TODO: logical vs computational *)
+       (* TODO: BaseType case *)
+       | _ ->
+         let f (s, _) = Sym.equal s v in
+         (match List.find_opt f iargs with
+          | Some _ -> Yes ([], var_cands)
+          | _ -> Unknown (!^"Could not find variable definition line for" ^^^ Sym.pp v)))
+    | Some line ->
```
dc-mak: Use a nested pattern-match here to reduce indentation.

backend/cn/lib/explain.ml
```diff
@@ -26,6 +26,430 @@ type log_entry =
 
 type log = log_entry list (* most recent first *)
 
+(** Infrastructure for checking if a countermodel satisfies a predicate **)
+module CheckPreds = struct
```
dc-mak: At around 400 lines, this module might be better placed in a stand-alone file?

dc-mak (changes requested):
Sorry to come in at this stage and request this many changes. I think most/all of these are relatively small but if that's not the case, make the small changes and make follow-up issues for the bigger ones and add them in separate PRs asap.

For future reference: I appreciate it can be difficult to batch new features into incremental small PRs, but that will lead to quicker reviews, easier to understand commits and smaller requests for changes. Splitting changes up like takes some getting used to and I'm happy to help with that.

backend/cn/lib/explain.ml
cassiatorczon: I see, for some reason I thought we were already using it. I'll remove that for now.

backend/cn/lib/solver.ml
cassiatorczon: I don't, no. I may be misunderstanding how the effectful code is working here; are there specific changes you'd recommend making if I never re-use `s`?

backend/cn/lib/definition.ml
cassiatorczon: Negative guards are passed to the solver.

backend/cn/lib/solver.ml
cassiatorczon: To clarify, is this comment the same as the other, just pointing out the argument too?

backend/cn/lib/solver.ml
cassiatorczon: This is actually @cp526 's code that I merged in because I was initially planning to use it (I didn't end up doing so). I have no stylistic preference here.

backend/cn/lib/solver.ml
cassiatorczon: @cp526 Is this needed for anything?

backend/cn/lib/explain.ml
cassiatorczon: This is here because we weren't sure if the UI would be easier to understand with or without this, so I may put this back later based on feedback once people have used it. Is it preferable for me to take the comment out entirely for now?
(I believe this is the case for all the code comments I left.)

backend/cn/lib/explain.ml
cassiatorczon: Same as above.

backend/cn/lib/explain.ml
cassiatorczon: Same as above.

backend/cn/lib/explain.ml
cassiatorczon: This is the same as the other code comments.

backend/cn/lib/solver.ml
cp526: If you swap the first two lines and put `debug_ack_command s (SMT.push 1);` before `let s' = assume_constraints lcs s in`, that's cleaner.

backend/cn/lib/solver.ml
cp526: I put it in for "watching" the new try-hard code: when that succeeds in proving something the regular mode failed to prove, it doesn't so far produce any other output, so it was useful in that regard.

backend/cn/lib/explain.ml
cp526: Sounds good, let's keep the comments.

backend/cn/lib/solver.ml
dc-mak: If you're not using the code, can you delete it?

backend/cn/lib/solver.ml
dc-mak: Correct

backend/cn/lib/explain.ml
dc-mak: Fair, please add a comment and an issue link in that case.

backend/cn/lib/solver.ml
cp526: We'll want to try using it quite soon (for context, @dc-mak : this is logic to make the SMT solver consider the full set of assumptions, including recursive definitions and quantifiers, in a "try-hard" mode).

So it wouldn't be bad to put it in now, I'd say, so we can get this merged quickly (also wrt to the upcoming repository split).

backend/cn/lib/explain.ml
cassiatorczon: Stylistically, do we prefer:
1. an assert (traverses the lists an extra time)
2. something like this, with an option returning Unknown in the None case
3. throwing some type of error (I see Cerberus has its own error handling mechanisms; is there a preferred one to use here)?

backend/cn/lib/explain.ml
cassiatorczon: I'm confused which part of the code this is referring to.

backend/cn/lib/explain.ml
cp526: In this case I'd probably just use regular `List.zip` (which can't fail if we know that the iarg lists have the same length), but I'm not too fussed about it.

cassiatorczon: I'm not seeing a regular `List.zip` in 4.14? https://ocaml.org/manual/4.14/api/List.html

cp526: Sorry, I think it's called List.combine.

backend/cn/lib/report.ml
cassiatorczon: I'm using the preexisting `label` type and infrastructure here - we can definitely change it to a datatype, but that would affect other people's code so I think it might be beyond the scope of this PR. I can added an issue for that. https://github.com/rems-project/cerberus/issues/899

backend/cn/lib/report.ml
cassiatorczon: Added https://github.com/rems-project/cerberus/issues/900

backend/cn/lib/explain.ml
cassiatorczon: https://github.com/rems-project/cerberus/issues/900

backend/cn/lib/explain.ml
cassiatorczon: https://github.com/rems-project/cerberus/issues/900

backend/cn/lib/explain.ml
cassiatorczon: https://github.com/rems-project/cerberus/issues/900

backend/cn/lib/explain.ml
cassiatorczon: https://github.com/rems-project/cerberus/issues/900

backend/cn/lib/explain.ml
cassiatorczon: https://github.com/rems-project/cerberus/issues/901

backend/cn/lib/solver.mli
cassiatorczon: I wanted to be able to 1) pass all definitions to the solver (now doable with `provable` and `try_hard`) and 2) be able to get back "Unknown" from the solver right now even with `try_hard` we always return `True` or `False`. I can extend that if you are okay with `provable` potentially returning "Unknown" in try-hard mode

cassiatorczon: I'm also a bit confused about all of the arguments to `provable`. I have a `Global.t` and a list of constraints that I want to check the consistency of. Christopher previously recommended adding all of the constraints to the solver as assumptions; should I still do that, or just `and` them together and make them the last argument? Would I leave the assumptions argument empty then? For `simp_ctxt`, can I follow the logic of `make_simp_ctxt` in typing.ml?

backend/cn/lib/explain.ml
cassiatorczon: Sort of. If the solver gives us back `[1;2]` as a countermodel that supposedly came from some predicate `P`, and the last line of `P` is `return x :: y`, then here `[1;2]` is the first candidate mentioned, and `1` and `[2]` are candidates for `x` and `y`, which we then recursively check against the definitions of those two.

backend/cn/lib/solver.ml
cassiatorczon: This is also @cp526 's code -- I changed it to a `List.fold_right` but will let him confirm that that makes sense here.

backend/cn/lib/explain.ml
cassiatorczon: The thing being iterated over is a `LAT.packing_ft` (structurally a linked list). I can define `fold_right` and `fold_left` for it if we'd like, but we couldn't use `List.fold_right` here.

backend/cn/lib/report.ml
cp526: String is just fine here. There really isn't any structure one would want to encode using a datatype.

cp526: We'll eventually want to rename `label` to something else, so as not to clash with Core (goto) labels, but not now.

backend/cn/lib/solver.ml
cp526: @dc-mak Turning this into a fold_right makes this strictly less readable.
@cassiatorczon Both versions are fine, go with whichever you have now.

backend/cn/lib/explain.ml
cp526: Let's leave it as is.

backend/cn/lib/solver.ml
cassiatorczon: After more testing, it looks like `try_hard` mode runs forever on some inputs. (I tested it on a minimal local copy I have of the `Queue` example, and it did on that.) I just deleted the `try_hard` code for now.

cp526: Could we leave the `try_hard` code in place, and just disable it? Otherwise the relationship between this PR and my PR that adds the try_hard logic gets too complicated: if we leave the try_hard code here, this PR will just subsume mine.

It's presumably Z3 that does not terminate on certain solver queries? That's also what I've observed, CVC5 works better, though presumably both can be made to work with the right settings.

cp526: ... Fully removing all the try hard code would also work, then I could merge my pull request after yours, but that only works if there's no overlap.

backend/cn/lib/solver.ml
dc-mak: This code should not be affected by the repo split, and if it is, I'm happy to take on the job of rebasing it.

backend/cn/lib/checkPredicates.ml
```diff
@@ -0,0 +1,415 @@
...
+(* The core function is `check_pred`, which, given a predicate and a term
+   matching the return type of the predicate, checks if the term is in the
+   range of that predicate. The term here is referred to as a "candidate".
+   The algorithm works by "walking backward" over the predicate: it
+   compares the candidate to the `return` expression of the predicate. If they
+   are compatible (e.g., use the same constructor), then this implies candidate
+   values for the free variables in the `return` expression.
+   (`get_var_cands` identifies these candidates.) These new
+   candidates are then recursively checked against the definitions of those
+   variables. *)
```
dc-mak, line 22: Thank you! Extremely helpful :-)

backend/cn/lib/checkPredicates.ml
```diff
@@ -0,0 +1,415 @@
...
+  (* Take the union of two symbol maps,
+     failing on any key that is in both maps with different values *)
+  let merge eq m1 m2 =
+    let comb k v acc =
+      let@ macc = acc in
+      match Sym.Map.find_opt k macc with
+      | Some v' -> if eq v v' then acc else No (Pp.( !^ ) "Incompatible list elements")
+      | None -> Yes (Sym.Map.add k v macc)
+    in
+    Sym.Map.fold comb m1 (Yes m2)
```
dc-mak, line 57: Do https://ocaml.org/manual/5.2/api/Map.Make.html#VALmerge or https://ocaml.org/manual/5.2/api/Map.Make.html#VALunion  do what you need here?

backend/cn/lib/resultWithData.ml
dc-mak, line 1: I'll make an issue to replace this file with ``('a, [`No | `Unknown | `Error] * 'b) Result.t``

backend/cn/lib/solver.mli
dc-mak: This are all very good questions and the honest answer is, I don't know, I've had many of the same questions myself at some point. I am quite happy with what you've done and the rest can be left for a technical-debt issue.

dc-mak (commented):
Looking good - couple of things, but happy to ship and address in a follow-up as well if pragmatic.

backend/cn/lib/resultWithData.ml
dc-mak: #903

backend/cn/lib/solver.mli
dc-mak: #902

dc-mak (approved):
Whoever squashes and merges this in, please make sure the commit message contains the top comment on this PR.

backend/cn/lib/solver.ml
cp526: Both, leaving the try-hard code in place, or removing it altogether, would be fine and should be quick to do. Just, we shouldn't have a mix where part of the try-hard code is still here and the rest is removed.

Either way, the try-hard code can be merged in before the split, that's the most efficient.

backend/cn/lib/checkPredicates.ml
cassiatorczon: I don't think so -- if the two maps contain different values for the same key, then I do not want a map at all as the output, I want it to fail with `No`

cp526:
It looks like it's now just the formatting that's failing. When that's fixed, and if you're happy, we can go ahead and merge. Probably we'll have to do the squash-and-merge option.

cassiatorczon:
Sorry, I should have left a comment - I was just debugging, one of the last set of changes seems to have introduced a bug. (The bug just seems to cause `checkPred` to miss some predicates it previously correctly identified as invalid. It doesn't show any erroneous information.)

cp526:
> Sorry, I should have left a comment - I was just debugging, one of the last set of changes seems to have introduced a bug. (The bug just seems to cause `checkPred` to miss some predicates it previously correctly identified as invalid. It doesn't show any erroneous information.)

Ah, sorry, I had assumed this was finished. In that case, best to make a small new PR with a fix (rather than `revert`ing committed changes, I think).

dc-mak:
> Whoever squashes and merges this in, please make sure the commit message contains the top comment on this PR.

I would very much like to improve our commit standards and cooperation on this would be much appreciated. If you have not reviewed this in a while, please take a moment to remind yourself of this. If you disagree, please open a separate issue and explain what and why.
https://github.com/rems-project/cerberus/blob/master/backend/cn/CONTRIBUTING.md#git-guidelines 

I have struggled with the uninformative messages and non-atomic commits often enough in this repo to have wasted a lot time and effort. On the other hand, when the commits are small and the messages accurate, though this meant a little bit of extra work for the original author, this helped tremendously with understanding, debugging, maintaining, updating, refactoring, catching regressions, and adding new features.

Having tests and reasonable commits, **including on development branches** helps avoid regressions and bugs.

PR #833, opened 2025-01-22
Makes cached model evaluator solver resettable

kiranandcode:
The CN LSP Server uses CN as a library and makes repeated calls to `Cn.Check.check_decls_lemmata_fun_specs` to verify and type check multiple files. As each verification call is intended to be in a fresh context, this seems to cause a problem with the cached solver used by the model evaluator and causes an error to be raised leading to the following issue: https://github.com/GaloisInc/VERSE-Toolchain/issues/142

This PR exposes a function in the Solver interface that allows resetting this state.

(The VERSE-toolchain uses an older commit of Cerberus, but applying this patch to that commit also makes the above error go away so it seems to be the cause of the problem)

### DIFF
--- a/backend/cn/lib/solver.ml
+++ b/backend/cn/lib/solver.ml
@@ -1325,17 +1325,22 @@ let model_state = ref No_model
 let model () = match !model_state with No_model -> assert false | Model mo -> mo
 
 (** Evaluate terms in the context of a model computed by the solver. *)
-let model_evaluator =
-  let model_evaluator_solver = ref None in
+let model_evaluator, reset_model_evaluator_state =
+  (* internal state for the model evaluator, reuses the solver across consecutive calls for efficiency *)
+  let model_evaluator_solver : Simple_smt.solver option ref = ref None in
   let currently_loaded_model = ref 0 in
-  let new_model_id =
-    let model_id = ref 0 in
-    fun () ->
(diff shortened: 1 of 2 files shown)


kiranandcode:
(I'm not familiar enough with CN to understand what exactly is causing this bug, this fix seems to do the trick, an alternative might be to properly reset the solver state between multiple calls (maybe using `(reset)`?))

samcowger:
@yav and @dsainati1, it looks like you both wrote some of the underlying code here, does this change seem appropriate to you, and/or do you have any thoughts on a `(reset)`-based approach?

yav:
This looks reasonable to me.   I wonder if instead of exporting `reset_solver`, we should just update `make`  to call `reset_solver`. 

Slightly longer run, I think it would make sense to update the API to have `model` take a `solver` as a parameter, and package all of a solver's state with it, rather than using these global variables (the current API is for compatibility with the all solver, but I think by now we've successfully transitioned to using external solvers, so we can just change it).

cp526:
Thanks @kiranandcode ! @yav Is this good to go, shall I merge?

yav:
The changes makes sense on their own, so we can merge it.    I do think it might make sense to also call `reset_solver` from `make` though, as I don't think it would ever make sense to call `make` twice, without also resetting the evaluation solver.

PR #855, opened 2025-02-07
seq tests

echoumcp1:
USAGE:
1. cd to the directory containing the file you want to test.
2. Suppose your file is called test.c. 
   Run the following command to generate a sequence of 100 random calls. "cn seq_test test.c --output-dir="test"
   The output-dir parameter is required. The call above places the resulting files in the test directory.
   If a nonexistent directory is passed in, the directory will be automatically created and the test files placed in there.
   There is an optional --num-samples parameter that specifies the number of calls to generate. 
   Some other useful options are --max-backtrack-attempts and --max-resets.
   --max-backtrack-attempts sets maximum the number of times to attempt generating a call.
   --max-resets sets the maximum number of times to clear the generation context (which records the previous calls
   for reuse in later calls).
3. After running "cn seq_test..." there will be a run_tests.sh script. Running the script will link all the object files, 
   create the executable as "tests.out" and run the executable.


runtime/libcn/include/.DS_Store
kmemarian, line 1: please remove this file before the merge

tests/cn-seq-test-gen/src/abs.pass.c
ZippeyKeys12, line 1: Are you going to add your tests to the spec testing workflow or make a separate one? (I don't know the tradeoff of having too many separate workflows)

backend/cn/lib/executable_spec.ml
```diff
@@ -381,6 +381,7 @@ let main
       let global_ownership_init_pair = generate_ownership_global_assignments sigm prog5 in
       global_ownership_init_pair @ executable_spec.pre_post)
   in
+  let _ = source_file_in_stmt_injs in ();
```
ZippeyKeys12: What is this line for?

backend/cn/bin/main.ml
```diff
@@ -1119,6 +1206,11 @@ module Testing_flags = struct
   let trap =
     let doc = "Raise SIGTRAP on test failure" in
     Arg.(value & flag & info [ "trap" ] ~doc)
+
+  let max_resets =
```
ZippeyKeys12: The flags are sorted by subcommand, so you should make a module (ex: `Seq_Testing_flags`) to add any unique to your subcommand.

backend/cn/lib/testGeneration/testGenConfig.ml
```diff
@@ -22,7 +22,8 @@ type t =
     sized_null : bool;
     coverage : bool;
     disable_passes : string list;
-    trap : bool
+    trap : bool;
+    max_resets : int
```
ZippeyKeys12: You should make your own config type for seq test specific stuff.

backend/cn/lib/testGeneration/testGeneration.ml
```diff
@@ -380,3 +380,16 @@ let run
...
+  ~filename
+  (sigma : Cerb_frontend.GenTypes.genTypeCategory Cerb_frontend.AilSyntax.sigma)
+  (prog5 : unit Mucore.file)
+  : unit
+  =
+  Cerb_debug.begin_csv_timing ();
+  if Option.is_some prog5.main then
+    failwith "Cannot test a file with a `main` function";
+  SeqTests.generate ~output_dir ~filename sigma prog5 ;
+  Cerb_debug.end_csv_timing "specification test generation"
```
ZippeyKeys12: Needs a different name, presumably something like "sequence testing"?

ZippeyKeys12 (changes requested):

tests/cn-seq-test-gen/src/abs.pass.c
echoumcp1: I'm not sure either, but I didn't copy all of the spec testing tests since the tests with functions that require pointers but have no way of generating them will not work with the current iteration of seq testing.

backend/cn/lib/executable_spec.ml
echoumcp1: I do not remember modifying this file at all, but that line just looks like it does nothing anyway.

backend/cn/lib/testGeneration/testGenConfig.ml
echoumcp1: Should I put put the seq config stuff in its own seqTestGenConfig or something along those lines?

ZippeyKeys12: The reason for `testGenConfig.ml` is because it is used across multiple files and was added later in development, so it was just easier. I think you could just put the config in `seqTests.ml` and pass it in as an argument (I plan to refactor my stuff to do something similar at some point).

Otherwise, I'd put it in a file like `seqTestConfig` or something, yeah.

backend/cn/lib/executable_spec.ml
ZippeyKeys12: Could you delete it then?

backend/cn/bin/main.ml
```diff
@@ -443,6 +443,93 @@ let generate_executable_specs
...
+            prog5
+            statement_locs;
+          let config : TestGeneration.seq_config = 
+              { 
+                num_samples = num_samples;
+                max_backtracks = backtrack_attempts;
+                num_resets = num_resets 
+              } in
+          TestGeneration.set_seq_config config;
+          let _ = statement_locs, cabs_tunit in
```
ZippeyKeys12: Delete this line?

backend/cn/lib/testGeneration/testGenConfig.mli
```diff
@@ -22,7 +22,7 @@ type t =
     sized_null : bool;
     coverage : bool;
     disable_passes : string list;
-    trap : bool
+    trap : bool;
```
ZippeyKeys12: You need to run ocamlformat

ZippeyKeys12:
Thinking about it, with Rini's upcoming refactor, `cn instrument` will have its own subdirectory. Already, `cn test` has its own (though currently this PR would break that). At some point this should be in its own subdirectory. It'll also fully separate your work, so you shouldn't run into any conflicts anymore. Whether you want to do this now or later is up to you (if not now make an issue).

tests/cn-seq-test-gen/src/abs.pass.c
ZippeyKeys12: If you just add your tests to CI, I'll merge this

echoumcp1: Is it okay to just duplicate the run-cn-test-gen.sh, modify it for seq testing, and create a new yaml or is there something else I should do?

ZippeyKeys12: For the shell script, sure. For the YAML file, It's up to you whether you make a new one or just add it as a new step to `ci-cn-spec-testing.yml`.

echoumcp1: The seq_test command seems to be returning 0 even if the test fails, but if I run the run_tests.sh script then it returns a nonzero exit code as expected for a failing test. Unfortunately this means the CI script registers a correctly failing test as a successful test. I could just run the run_tests.sh again and use that exit code within the script, but I think cn seq_test should behave the same as cn test w.r.t the exit code. Where are you propagating the exit code from run_tests.sh in cn test?

ZippeyKeys12: In `run_seq` you'd have to save the exit code and when it's a failure call `exit <code>`. I don't have to do this; `execv` replaces the existing process.

echoumcp1: Got it. Also, running the seq_test with static hack causes a number of duplicate symbol issues with the linker (ex. runway.pass.c case). I'll make an issue for this.

ZippeyKeys12 (approved):

ZippeyKeys12:
[Bug involving pointer types](https://github.com/rems-project/cerberus/actions/runs/13383870299/job/37408221927?pr=855#step:13:3518)

PR #857, opened 2025-02-08
Fix docker build

podhrmic:
Update the build based on the latest documentation, and build the docker image for users (no dev dependencies needed).


dc-mak:
What's the status on this? I keep getting notifications about the failing CI.

podhrmic:
See #873 - I will open a new PR, as this branch does not have the right fix

PR #847, opened 2025-02-04
CN: Mucore program export to Coq

vzaliva:
This PR introduces the following functionality to CN:

1. Coq definitions for MuCore and related Cerberus and CN types (in `backend/cn/coq/MuCore/`).
2. A command-line option for `cn verify`: `--coq-export-file=`, which generates and exports _mucore_ as a Coq file.
3. A script to typecheck the generated files: `tests/run-cn-coq.sh`.
4. CI integration.

(When merged, all commits should be squashed.)

PR #864, opened 2025-02-12
[CI] Trying to fix CI

ZippeyKeys12:


ZippeyKeys12:
I think #847 passed, but broke later CI, because the errors came from caching?

ZippeyKeys12:
There's something happening with the caching, but I evidently didn't fix it

PR #865, opened 2025-02-12
[CI] Trying to CI take 2

ZippeyKeys12:
Follow-up to #864


ZippeyKeys12:
There's two errors, one is that for PRs the cache does not have the CHERI switch, the other is that the base CI is trying to create a switch that already exist, when the cache doesn't exist.

ZippeyKeys12:
I think this might fix the second, but it's unclear how to tell without merging it...

ZippeyKeys12:
Okay, the solution to the first is rebasing on `master` and deleting the old [cache](https://github.com/rems-project/cerberus/actions/caches) for your PR.

Issue #823, opened 2025-01-08
Preprocessor and executable specifications


yav:
There are some unfortunate interactions between the C pre-processor and CN's testing infrastructure (e.g., see #393, #654).
This ticket aims to describe the underlying problem and some potential design choices that we may want to consider.

Status Quo
----------------

The following diagram shows the steps that CN takes to produce executable specifications:
```mermaid
graph LR
  input.c --> |cpp| input.i
  input.i --> |parser| AIL
  AIL --> |semantics| core
  core --> |synthesize| checks
  checks --> |inject| output.c
  input.c --> |inject| output.c
```
Here `AIL` is a data structure, which is an abstract representation of C code, and the `checks` is a bunch of C code (currently represented as a mixture of `AIL` and raw strings) that needs to be added to the original program to validate user assertions.  The checks are added to the original program by the `inject` algorithm.

Problems and Considerations
--------------------------------------------

**Incorrect Column Numbers**  Note that because the `parser` runs after the pre-processor has been applied, the location information we have is not quite right.    We are using an external pre-processor, which annotates its output with file+line information, however *column* information is lost, so the locations that are stored in the AIL have correct collumns for `input.i` but *NOT* for `input.c`

The effect of this is two-fold:
  1. Error messages may contain location information where the column is wrong
  2. The `inject` algorithm gets confused, resulting in incorrect `ouput.c`

**Multiple Files** Because `input.c` may contain `#include` directives, when we perform injection we may have to perform injection in *multiple* files.   We do have some code that attempts to do this, but it doesn't seem to work quite correctly.  This is a tricky problem, because generally we don't want to modify the original files, but rather a copy of the files.  We then have to ensure that those modified files are correctly included in `output.c` when building it, but we have no easy way to modify the `#include`s, so we have to achieve it only by changing the flags to the pre-processor.   In principle this will never be fully correct, for example, if a file `#included` and absolute path, we can't change that.   Fortunately, this is not something that usually happens in normal programs.

Potential Solutions
---------------------------

Here are a few possibilities we may want to consider:

(A) Modify the `inejct` algorithm to be line-based and not consider columns.   This will only work if the injections can always go on new lines, which---in general---is not the case.   Consider, for example, a program that has multiple statements on the same line:  we may have to inject stuff in-between those statements.

[...]

PR #866, opened 2025-02-12
Track location in pre-processed file, as well references to source.

yav:
This is the first step of solving #823.

We introduce a new module `Cerb_location.position`, which defines a type replacing `Lexing.position` as the base type to use for positions. Values of the new position type keep the `Lexing.position` corresponding to the actual location in the preprocessed file, as well as the file name and line in the original source associated with the location.

### DIFF
--- a/backend/cn/bin/main.ml
+++ b/backend/cn/bin/main.ml
@@ -43,7 +43,14 @@ end
 
 open Log
 
-let frontend ~macros ~incl_dirs ~incl_files astprints ~filename ~magic_comment_char_dollar
+let frontend
+  ~macros
+  ~incl_dirs
+  ~incl_files
+  astprints
+  ~filename
+  ~magic_comment_char_dollar
+  ~save_cpp
   =
   let open CF in
   Cerb_global.set_cerb_conf
@@ -66,7 +73,7 @@ let frontend ~macros ~incl_dirs ~incl_files astprints ~filename ~magic_comment_c
      @ if magic_comment_char_dollar then [ "magic_comment_char_dollar" ] else []);
   let@ stdlib = load_core_stdlib () in
   let@ impl = load_core_impl stdlib impl_name in
-  let conf = Setup.conf macros incl_dirs incl_files astprints in
+  let conf = Setup.conf macros incl_dirs incl_files astprints save_cpp in
   let cn_init_scope : Cn_desugaring.init_scope =
     { predicates = [ Alloc.Predicate.(str, sym, Some loc) ];
       functions = List.map (fun (str, sym) -> (str, sym, None)) cn_builtin_fun_names;
@@ -132,6 +139,7 @@ let with_well_formedness_check
   ~astprints
   ~no_inherit_loc
   ~magic_comment_char_dollar
+  ~save_cpp
   ~(* Callbacks *)
    handle_error
   ~(f :
@@ -151,7 +159,8 @@ let with_well_formedness_check
          ~incl_files
          astprints
          ~filename
-         ~magic_comment_char_dollar)
+         ~magic_comment_char_dollar
+         ~save_cpp)
   in
   Cerb_debug.maybe_open_csv_timing_file ();
   Pp.maybe_open_times_channel
@@ -265,6 +274,7 @@ let well_formed
     ~astprints
     ~no_inherit_loc
     ~magic_comment_char_dollar
+    ~save_cpp:None
     ~handle_error:(handle_type_error ~json ?output_dir ~serialize_json:json_trace)
     ~f:(fun ~cabs_tunit:_ ~prog5:_ ~ail_prog:_ ~statement_locs:_ ~paused:_ ->
       Or_TypeError.return ())
@@ -345,6 +355,7 @@ let verify
     ~astprints
     ~no_inherit_loc
     ~magic_comment_char_dollar
+    ~save_cpp:None
     ~handle_error:(handle_type_error ~json ?output_dir ~serialize_json:json_trace)
     ~f:(fun ~cabs_tunit:_ ~prog5:_ ~ail_prog:_ ~statement_locs:_ ~paused ->
       let check (functions, global_var_constraints, lemmas) =
@@ -377,6 +388,11 @@ let handle_error_with_user_guidance ~(label : string) (e : exn) : unit =
   Printf.eprintf "Prefix your issue with \"[%s]\". " label;
   Printf.eprintf "Check that there isn't already one for this error.\n";
   exit 1
+
+
+let pick_cpp_file_name outdir filename =
+  let cpp_name = Filename.remove_extension filename ^ "-preproc.c" in
+  match outdir with None -> cpp_name | Some d -> Filename.concat d cpp_name
 
 
 let generate_executable_specs
@@ -429,6 +445,15 @@ let generate_executable_specs
   Diagnostics.diag_string := diag;
   WellTyped.use_ity := not no_use_ity;
   Sym.executable_spec_enabled := true;
+  (* XXX temporary: should we inject in the pre-processed file or original one *)
+  let use_preproc = false in
+  let fi, save =
+    if use_preproc then (
+      let fi = pick_cpp_file_name output_dir filename in
+      (fi, Some fi))
+    else
+      (filename, None)
+  in
   with_well_formedness_check (* CLI arguments *)
     ~filename
     ~macros
@@ -441,6 +466,7 @@ let generate_executable_specs
     ~astprints
     ~no_inherit_loc
     ~magic_comment_char_dollar (* Callbacks *)
+    ~save_cpp:save
     ~handle_error:(handle_type_error ~json ?output_dir ~serialize_json:json_trace)
     ~f:(fun ~cabs_tunit:_ ~prog5 ~ail_prog ~statement_locs ~paused:_ ->
       Cerb_colour.without_colour
@@ -452,7 +478,8 @@ let generate_executable_specs
                ~with_loop_leak_checks
                ~with_test_gen
                ~copy_source_dir
-               filename
+               fi
+               ~use_preproc
                ail_prog
                output_decorated
                output_decorated_dir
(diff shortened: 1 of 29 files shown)


yav:
@kmemarian @cp526 this PR contains the changes to the lexer and parser to keep track of both the position in the pre-processed file, and also the position in the original source.   Please have a look and let me know what you think.

At the moment the position in the actual file is not really used, but the next step would be to modify the injection code
to use to inject code in the pre-processed file.

util/cerb_position.ml
```diff
@@ -0,0 +1,66 @@
...
+let dummy = { file = Lexing.dummy_pos; source_file = ""; source_line = 0 }
+
+let from_lexing p =
+  { file = p; source_file = p.pos_fname; source_line = p.pos_lnum; }
+
+let line pos = pos.source_line
+let file pos = pos.source_file
+
+
+(* Column for this position, 0 based *)
```
kmemarian: this should say `1 based`

backend/cn/lib/locations.ml
```diff
@@ -59,12 +59,11 @@ let unpack l = l
...
-let json_lexing_position p =
-  let open Lexing in
+let json_lexing_position pos =
   `Assoc
-    [ ("file", `String p.pos_fname);
-      ("line", `Int p.pos_lnum);
-      ("char", `Int (p.pos_cnum - p.pos_bol))
+    [ ("file", `String (Cerb_position.file pos));
+      ("line", `Int (Cerb_position.line pos));
+      ("char", `Int (Cerb_position.column pos))
```
kmemarian: It's probably not an issue, but note that since `Cerb_position.column` is 1-based (which I think is reasonable) this diff is observable.

util/cerb_position.mli
kmemarian, line 1: this all look good, modulo a few typos in the comments.

parsers/c/c_parser_driver.ml
```diff
@@ -63,8 +64,9 @@ let diagnostic_get_tokens ~inside_cn loc string =
         let Lexing.{ pos_lnum; pos_bol; pos_cnum; _ } = lexbuf.lex_start_p in
         let (line, col) =
           (* the first line needs to have columns shifted by /*@ but the rest do not *)
-          let col_off = if pos_lnum > 1 then 1 else start_pos.pos_cnum - start_pos.pos_bol + 1 in
-          (pos_lnum + start_pos.pos_lnum, col_off + pos_cnum - pos_bol) in
+          let col_off = if pos_lnum > 1 then 1 else Cerb_position.column start_pos + 1 in
```
kmemarian: Is the `+ 1` right? This makes it different from the previous version.

kmemarian (commented):
Thanks for working on this, all looks fine apart from the off-by-one difference in `c_parser_driver.ml` (see inline comment).

Regarding the new global state in the lexer, it would indeed be better without it but I also don't see how we can do better for now.

yav:
Thanks, I've fixed these issues, and a bunch of typos in the comments.

kmemarian (approved):

kmemarian:
@yav Looks good to me, shall we merge?

yav:
Thanks, merged!

backend/cn/bin/main.ml
```diff
@@ -429,6 +445,15 @@ let generate_executable_specs
   Diagnostics.diag_string := diag;
   WellTyped.use_ity := not no_use_ity;
   Sym.executable_spec_enabled := true;
+  (* XXX temporary: should we inject in the pre-processed file or original one *)
```
dc-mak, line 448: Please use your initials instead of XXX, and make an issue.

backend/cn/bin/main.ml
```diff
@@ -43,7 +43,14 @@ end
 
 open Log
 
-let frontend ~macros ~incl_dirs ~incl_files astprints ~filename ~magic_comment_char_dollar
+let frontend
+  ~macros
+  ~incl_dirs
+  ~incl_files
+  astprints
```
dc-mak, line 50: Is there a good reason this is not named when `astprints` in other functions is?

backend/cn/lib/cStatements.ml
```diff
@@ -6,7 +6,9 @@ module LocCompare = struct
 
   type pre_cmp = int * (int * int * int * string) list * string list [@@deriving eq, ord]
 
-  let lex_to_cmp l =
+  let lex_to_cmp pos =
+    let l = Cerb_position.to_file_lexing pos in
+    (* XXX: file or source location? not sure why we are compring positions here, *)
```
dc-mak, line 11: Initials and issue here please.

backend/cn/lib/pp_mucore_coq.ml
```diff
@@ -451,7 +451,11 @@ let pp_memory_order = function
 
 let pp_polarity = function Core.Pos -> !^"Pos" | Core.Neg -> !^"Neg"
 
-let pp_lexing_position { Lexing.pos_fname; pos_lnum; pos_bol; pos_cnum } =
+(* XXX: use the original source locations? *)
```
dc-mak, line 454: Initials and issue here please too.

parsers/c/c_lexer.mll
```diff
@@ -7,16 +7,55 @@ open Tokens
...
+  let get () = !mapping
+
+  (** Make a position from a Lexing.position, consulting the line map *)
+  let position x =
+    let p1 = Cerb_position.from_lexing x in
+    let line_map = !mapping in
+    let src_loc = Cerb_position.LineMap.lookup (Cerb_position.line p1) line_map in
+    Cerb_position.set_source src_loc p1
+
+end
```
dc-mak, line 47: This is definitely possible, and only looks like few lines of code - would you like to try this in a follow-up PR?
https://schu.be/til-sedlex-and-menhir.html
https://gitlab.inria.fr/fpottier/menhir/blob/master/lib/Convert.mli#L69

parsers/c/c_parser_driver.mli
```diff
@@ -0,0 +1,41 @@
+
+(** Run only the lexer on a given string.
+The result is (token string, (line,column))
+XXX: It might be nice to generalize this to show both source and file locs.
```
dc-mak, line 4: Initials and issue.

parsers/c/c_parser.mly
dc-mak, line 1: This set of changes is a combination of a refactor, and a semantics change, which should be separated into different PRs.

backend/cn/lib/source_injection.ml
```diff
@@ -422,37 +441,39 @@ let output_injections oc cn_inj =
...
-    | AilSswitch (_, s)
-    | AilScase (_, s)
-    | AilScase_rangeGNU (_, _, s)
-    | AilSdefault s
-    | AilSlabel (_, s, _)
-    | AilSmarker (_, s) ->
-      aux acc s
-  in
-  aux [] stmt
+(* This appears to be unused:
```
dc-mak, line 444: Delete if so (in a separate PR).

parsers/c/c_parser_driver.mli
dc-mak, line 1: Thanks for creating and documenting!

parsers/core/core_parser.mly
dc-mak, line 1: Same as before - refactoring and semantic changes should be separated.

dc-mak (commented):
Thanks for the extra documentation comments in the code! And for doing this fiddly piece of work, and getting this merged in. If you are able to address the comments in follow-up PRs, that would be much appreciated!

Also if the commit messages in the future could follow the guidelines stated here, that would be great too: https://cbea.ms/git-commit/#seven-rules

Specifically, 50 chars for the header and a short explanation.

dc-mak:
> Thanks for the extra documentation comments in the code! And for doing this fiddly piece of work, and getting this merged in. If you are able to address the comments in follow-up PRs, that would be much appreciated!

backend/cn/bin/main.ml
yav: Having a consistent way to mark potential issues with the code is more useful than knowing who wrote the comment, because it enables everyone to quickly locate problems with the code.

I'd be happy to add my initials to this, but if one wanted to know who added the comment, they could always use `git blame` to look it up.

backend/cn/bin/main.ml
yav: I didn't add this flag, I just add the `cpp` one.   This shows up as a change because of the requirements to run the auto-formatter, which added extra changes.

parsers/c/c_parser.mly
yav: Can you say a bit more on what you mean here?   In particular what you consider to be a semantic change vs. refactoring?

 The only semantic change I intended to do was keep both positions, and I did it by adding "smart" constructor that consult the line map when constructing things.  I guess one could consider this a refactor, but it seemed like the only manageable way to do the change.

If I changed something else, that's unintentional and we should certainly review and probably revert.

backend/cn/lib/source_injection.ml
yav: In principle I also like deleting unused stuff which is unused, but since this code is in development, I wasn't sure if this is perhaps something that we intended to use, but is not yet plugged in.

parsers/c/c_lexer.mll
yav: Ah, nice!  I'll certainly try to do this, but I think we should wait until the repository split.

backend/cn/bin/main.ml
dc-mak: In that case, we use TODO or FIXME in other places

backend/cn/lib/source_injection.ml
dc-mak: Fair, my MO is delete by default unless someone objects with good reason.

parsers/c/c_parser.mly
dc-mak: I would consider the addition of the smart constructor as one change (a refactor) and the addition of "both positions" as the second change (feature).

PR #878, opened 2025-02-18
Resource inference logging (with Coq export)

vzaliva:
- Log resource inference steps (just `predicate_request` for now)
- Log export in Coq formal is enabled by the `--coq-proof-log` option (must be used together with `--coq-export-file`)
- Coq classes for the proof log and related CN types
- Split Coq definitions into `Cn` and `Cerberus` packages.
- Enhanced cn-coq unit testing script with additional options.
- Export resource inference logs to Coq and type-check the results in CI. 

(this PR contains multiple commits, which should be squashed on merge).


vzaliva:
hold on with this PR. I want to do another tiny change today: do not collect log when it is not enabled via command line. It is quite big and eats lot of memory.

vzaliva:
it is ready now

backend/cn/bin/main.ml
```diff
@@ -507,6 +521,7 @@ let run_tests
     ~incl_files
     ~csv_times
     ~coq_export_file:None
+    ~coq_proof_log:false
     ~log_times
```
cp526, line 614: As far as I can see, only the verify function should need the Rocq argument -- not these and other testing-related functions.

backend/cn/lib/explain.mli
```diff
@@ -7,7 +7,6 @@ type action =
   | Call of Sym.t * IndexTerms.t list
   | Return of IndexTerms.t
 
-(** Info about what happened *)
```
cp526: Why drop the comment?

cp526 (commented):
Other than the above, LGTM

backend/cn/lib/explain.mli
vzaliva: it was removed by mistake. Fixed now.

backend/cn/bin/main.ml
vzaliva: Due to this line:

`let subcommands = [ wf_cmd; verify_cmd; testing_cmd; instrument_cmd ]`

All sub-commands functions expect to have the same signature.

vzaliva: To clarify: mpst commands re-use `with_well_formdness_check` function which takes coq-related flags. In this case for `run_seq_tests` command, the command itself does not take any coq-related flags, and passes `None` and `false` values to  `with_well_formdness_check` to prevent proof log generation.

vzaliva:
@cp526 The PR has been re-based and is ready now for another review and merge.
Zain A. confirmed that this "CN Spec Testing" CI failure is due to a bug unrelated to this PR.
All other CI tests pass.

PR #884, opened 2025-02-21
[CN-exec] Runtime loop invariants

rbanerjee20:
Implementation of runtime loop invariant checking within Fulminate

### DIFF
--- a/backend/cn/bin/main.ml
+++ b/backend/cn/bin/main.ml
@@ -527,7 +527,7 @@ let run_seq_tests
           let _, sigma = ail_prog in
           Cn_internal_to_ail.augment_record_map (BaseTypes.Record []);
           Executable_spec.main
-            ~without_ownership_checking (* TODO: Zain *)
+            ~without_ownership_checking (* TODO: Ethan *)
             ~without_loop_invariants:true
             ~with_test_gen:true
             ~copy_source_dir:false
@@ -667,7 +667,7 @@ let run_tests
           Cn_internal_to_ail.augment_record_map (BaseTypes.Record []);
           (try
              Executable_spec.main
-               ~without_ownership_checking (* TODO: Zain *)
+               ~without_ownership_checking (* TODO: Ethan *)
                ~without_loop_invariants:true
                ~with_test_gen:true
                ~copy_source_dir:false
--- a/backend/cn/bin/main.ml
+++ b/backend/cn/bin/main.ml
@@ -406,6 +406,7 @@ let generate_executable_specs
   output_decorated_dir
   without_ownership_checking
   without_loop_invariants
+  with_loop_leak_checks
   with_test_gen
   copy_source_dir
   =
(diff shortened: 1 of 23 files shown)


backend/cn/bin/main.ml
```diff
@@ -527,8 +527,8 @@ let run_seq_tests
           let _, sigma = ail_prog in
           Cn_internal_to_ail.augment_record_map (BaseTypes.Record []);
           Executable_spec.main
-            ~without_ownership_checking
-            ~without_loop_invariants
+            ~without_ownership_checking (* TODO: Zain *)
```
ZippeyKeys12: This work is @echoumcp1's

bcpierce00:
Question about defaults: If I run with instrumentation on a file with no annotations, will I get dinged for missing loop invariants?  I know there was some discussion about this a while back, but I'm not sure where we settled.  (FWIW, I think it's better for loop invariant checking to be off by default, to provide a smoother onramp for new users, testing tutorial readers, etc.)

cp526:
> Question about defaults: If I run with instrumentation on a file with no annotations, will I get dinged for missing loop invariants? I know there was some discussion about this a while back, but I'm not sure where we settled. (FWIW, I think it's better for loop invariant checking to be off by default, to provide a smoother onramp for new users, testing tutorial readers, etc.)

Agreed. Ideally the specifying of a loop invariant itself is what turns the invariant checking on locally, just for that invariant.

bcpierce00:
Yes, that’s even better than a global default.

rbanerjee20:
Thanks @bcpierce00 @cp526 for your comments. I've now added some functionality for keeping track of whether the user has provided any CN loop spec. If they have, the loop gets instrumented with runtime checks; otherwise, no checks get inserted. Previously, the statements that CN generates internally whenever it sees a loop were also being translated and inserted by Fulminate. I agree that it doesn't make sense to make the output verbose with these checks if the user has not written any spec for their loop.

@cp526 if you could take a look at [~~this commit~~](https://github.com/rems-project/cerberus/pull/884/commits/02c7f24e8d46efdc80c86301eea27e2c45401f10) [this set of commits](https://github.com/rems-project/cerberus/pull/884/files/abef8954928710dd31aaf6e7d1a3bda7c70b63ed..6df1e33968c7942d98e5e65116a115f41442d540), that would be good, as I made a small change to `Mucore.Label` and had to propagate this throughout the CN codebase. If there is some better/easier way of keeping track of whether the user has provided any loop annotations, please let me know.

backend/cn/bin/main.ml
```diff
@@ -662,7 +670,9 @@ let run_tests
           Cn_internal_to_ail.augment_record_map (BaseTypes.Record []);
           (try
              Executable_spec.main
-               ~without_ownership_checking
+               ~without_ownership_checking (* TODO: Ethan *)
```
ZippeyKeys12: There are two subcommands, `cn test` (`run_tests`) and `cn seq_test` (`run_seq_tests`), the former worked on by me, the latter by @echoumcp1.

Rather than these comments, could you just make (separate) issues for these?
A TODO with a name doesn't give a lot of info...

PR #892, opened 2025-02-27
CN: Support accesses in specifications (fix #371)

### LINKED ISSUE #371: [CN] specs on declarations have a different syntax from specs on definitions
In particular `accesses` does not work but `trusted` also does not.

```
int x;

void f(void)
/*@ accesses x; @*/
{

}

void g(void);
/*@ spec g();
    accesses x;
@*/
```
```
% cn accesses_spec.c
accesses_spec.c:10:5: error: unexpected token after ';' and before 'accesses'
parsing "cn_fun_spec": seen "CN_SPEC LNAME VARIABLE LPAREN cn_args RPAREN SEMICOLON", expecting "CN_REQUIRES nonempty_list(condition) CN_ENSURES nonempty_list(condition)"
    accesses x;
    ^~~~~~~~ 
```

Looking at the grammar in the documentation, it looks like a declaration spec should refer to `function_spec` rather than hardcode only requires and ensures.

dc-mak:
DO NOT MERGE: temporary commit to share and test.

This commit unifies the handling of definition and declaration specifications. Previously, decl specs did not support accesses, cn_function, or trusted. The requires and ensures clauses they did support was desugared in the frontend Lem code.

This commit moves the desugaring of the decl specs to backend/cn, specifically Core_to_mucore.normalise_fun_map_decl, alongside the definition specs parsing and desugaring. It also moves the logic for combining multiple function specs into the same module, outside of Parse.

Extra special care needs to be taken in figuring out which desugaring state to use. I'm not exactly sure about this bit, but it seems to be working (I think thanks to the add_spec_arg_renames).

### DIFF
--- a/ocaml_frontend/cn_ocaml.ml
+++ b/ocaml_frontend/cn_ocaml.ml
@@ -358,12 +358,37 @@ module MakePp (Conf: PP_CN) = struct
             ; Dnode (pp_ctor "[CN]ensures", List.map dtree_of_cn_condition lmma.cn_lemma_ensures)
             ] ) 
 
+  let dtrees_of_opt_cn_conds = function
+    | None -> []
+    | Some (_loc, conds) -> List.map dtree_of_cn_condition conds
+
+  let dtrees_of_cn_func_spec
+      { cn_func_trusted
+      ; cn_func_acc_func
+      ; cn_func_requires
+      ; cn_func_ensures } =
+    let trusted =
+      Option.fold
+        cn_func_trusted
+        ~none:[]
+        ~some:(fun _ -> [ Dleaf (pp_ctor "[CN]trusted") ]) in
+    let acc_func =
+      match cn_func_acc_func with
+        | None -> []
+        | Some (_, CN_accesses globals) ->
+          [ Dnode (pp_ctor "[CN]accesses", List.map (fun x -> Dleaf (Conf.pp_ident x)) globals) ]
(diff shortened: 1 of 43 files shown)


ocaml_frontend/cn_ocaml.ml
```diff
@@ -359,10 +359,9 @@ module MakePp (Conf: PP_CN) = struct
             ] ) 
 
   let dtree_of_cn_spec s =
-    Dnode ( pp_ctor "[CN]spec of" ^^^ P.squotes (Conf.pp_ident s.cn_spec_name)
-          , [ Dnode (pp_ctor "[CN]args", dtrees_of_args Conf.pp_ident s.cn_spec_args)
-            ; Dnode (pp_ctor "[CN]requires", List.map dtree_of_cn_condition s.cn_spec_requires)
-            ; Dnode (pp_ctor "[CN]ensures", List.map dtree_of_cn_condition s.cn_spec_ensures)
+    Dnode ( pp_ctor "[CN]spec of" ^^^ P.squotes (Conf.pp_ident s.cn_decl_name)
+          , [ Dnode (pp_ctor "[CN]args", dtrees_of_args Conf.pp_ident s.cn_decl_args)
+            ; Dnode (pp_ctor "[CN]clauses", [] (*FIXME*))
```
dc-mak: Fill this in

dc-mak:
~~Also need to do tests properly~~ done

PR #896, opened 2025-03-01
Makefile: replace `which` with `command -v` (#895)

### LINKED ISSUE #895: Replace `which` with `command -v` in `Makefile`
At the top of the makefile:
```make
# Checking for required tools.
ifeq (,$(shell which dune 2> /dev/null))
$(error "Compilation requires [dune].")
endif
ifeq (,$(shell which lem 2> /dev/null))
$(error "Compilation requires [lem].")
endif
```
This is problematic when opam uses bwrap / `.opam/opam-init/hooks/sandbox.sh` on some distros (namely nixos), unable to find `which`. 

Equivalent functionality would be: 
```make
ifeq (,$(shell command -v dune 2> /dev/null))
$(error "Compilation requires [dune].")
endif
ifeq (,$(shell command -v lem 2> /dev/null))
$(error "Compilation requires [lem].")
endif
```

quinn-dougherty:
Preserves functionality completely, but allows it to be built on some slightly exotic distros that have a different relationship with sandboxing than current assumptions. 

Closes #895 

See threads: 
- [coq discourse](https://coq.discourse.group/t/refinedc-install-make-under-opam-cant-find-dune/2544)
- [opam issues](https://github.com/ocaml/opam/issues/6405)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,8 @@
 # Checking for required tools.
-ifeq (,$(shell which dune 2> /dev/null))
+ifeq (,$(shell command -v dune 2> /dev/null))
 $(error "Compilation requires [dune].")
 endif
-ifeq (,$(shell which lem 2> /dev/null))
+ifeq (,$(shell command -v lem 2> /dev/null))
 $(error "Compilation requires [lem].")
 endif
 


kmemarian:
Thank you for the PR, this is now merged. I see that this is for RefinedC, which at the moment fixes its dependency to an old commit so this won't yet fix the issue.

quinn-dougherty:
That's right. I'm just pinning our fork for now https://github.com/Beneficial-AI-Foundation/cerberus/tree/refinedc-packaging

PR #915, opened 2025-03-06
CN: Use mem_vip backend (fix #850)

### LINKED ISSUE #850: CN: VIP vs concrete memory model
Currently, CN instantiates the _concrete_ memory model instead, while in formal modelling, it's logic in Coq we want to use VIP.

This causes problems with exporting integer and pointer values to Coq ( https://github.com/rems-project/cerberus/pull/847 ), leading to an ugly workaround in which, in pretty-printing methods delegated to the memory model, we attempt to print VIP memory model definitions from the Concrete memory model implementation.

Perhaps we can try to switch CN to instantiate the VIP memory model instead?

dc-mak:


kmemarian:
This needs a few tweaks to `vip/common.ml` to sync up with the `concrete` and `ocaml_implementation`. Which I can do shortly.

kmemarian (approved):
Apart from the cn-coq failure that I don't understand (@vzaliva ?), this looks good to me.

vzaliva:
To fix Coq part, we need to adjust `pp_pointer_value_for_coq` to print constructors with `Mem` prefix. E.g:

   | PVnull -> !^"PVnull"

should be

   | PVnull -> !^"Mem.PVnull"

That should fix CI error.

dc-mak:
Thanks for the team effort! @kmemarian @vzaliva if good to go feel free to squash and merge in.

PR #923, opened 2025-03-08
Support magic comments in a function call

dc-mak:
This commit adds support in the parser, Cabs, untyped Ail, typed Ail and Core to having a single magic comment at the end of an call expression. To check it works on a simple example, it also updates some of the associated AST printing functions. At the Ail level, the existing `AnnotatedExpression` had a `list string` argument for standard-related annotations, and so, to accommodate both attributes (used to carry magic comments) and this, the `AnnotatedExpression` now has a `list Annot.annotation` rather than `Annot.attrs` as is the case for `AnnotatedStatement`.

```c
void foo() { }

int main()
{
    foo(/*@ arg @*/);
    return 0;
}
```


dc-mak:
@kmemarian I am getting some strange behaviour which I don't understand. Cabs has and prints the magic comment, Ail has and prints it at debug level >= 5 but not below, and Core never has it.

```c
void foo() { }

int main()
{
    foo(/*@ arg @*/);
    return 0;
}
```
| Status | command
|--------|-----------------------------------------------------------------|
| Works | `cerberus tmp.c --switches=at_magic_comments --ast core` |
|       | `cerberus tmp.c --switches=at_magic_comments --ast ail -d 5` |
| Fails | `cerberus tmp.c --switches=at_magic_comments --ast ail -d 4` |
|       | `cerberus tmp.c --switches=at_magic_comments --ast core -d 5` |

kmemarian:
When `debug_level <= 4` it is the type annotated Ail that is pprinted (see `Pipeline.c_frontend`). Maybe `GenTyping` is losing the attributes? This would explain why Core never has them, but then how is CN getting them?

dc-mak:
That was exactly the info I needed! I am adding _new_ attributes and annotations, inside function calls, rather than relying on existing ones, and so this path was never tested before (until now).

kmemarian (approved):

PR #935, opened 2025-03-12
CN: Use member_shift<ctype> syntax (fix #728)

### LINKED ISSUE #728: [CN] Allow `member_shift<ctype>` as a syntax
Current one is confusing. Stop-gap for #349 . Should be a very small change to the parser, making it more similar to `array_shift`
https://github.com/rems-project/cerberus/blob/ede24dc5b3409e7554279a6a919975e838530d2f/parsers/c/c_parser.mly#L1860-L1871

dc-mak:


dc-mak:
Fyi this is also a (smallish) breaking change for CN.

PR #937, opened 2025-03-17
Use menhir 3.0 in dune-project (fix #927)

### LINKED ISSUE #927: Use menhir 3.0 in dune-project
https://dune.readthedocs.io/en/stable/reference/dune/menhir.html

"Note that starting in version 3.0 of the Menhir extension, the .conflicts file is generated by default."

This is handy for explaining parser conflicts, which need to be extracted manually right now. However, this requires updating the dune-lang version too.

dc-mak:
This is to get a .conflicts file produced by default for parser conflicts.

### DIFF
--- a/cerberus.opam
+++ b/cerberus.opam
@@ -20,7 +20,7 @@ bug-reports: "https://github.com/rems-project/cerberus/issues"
 depends: [
   "cerberus-lib"
   "ocaml" {>= "4.12.0"}
-  "dune" {(>= "3.8.0" & < "3.13.0") | (>= "3.15.0")}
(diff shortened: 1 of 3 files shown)


kmemarian:
You need to also update the dune dependency in the .opam files (I guess to `"dune" {>= "3.15.0"}`).

dc-mak:
Wait do you? Is the dune language version tied to the dune version somehow?

kmemarian:
It looks like it. I just tried a `dune-project` with `(lang dune 3.13)` with `dune.3.12.2` and it wasn't happy. FYI, we skip `3.14.0` because there is a bug that affects us.

PR #952, opened 2025-04-29
Fix `Pp_ail` when returning function pointers

ZippeyKeys12:
Necessary for https://github.com/rems-project/cn/pull/111


kmemarian:
It seems to still not be quite right:
```c
void (*(*func2(int))(int, short))(int, char);
```
gets printed back to:
```c
void (*) (signed int, char) (* func2(signed int)) (signed int, signed short); // Ail pprint
```

do you want to fix that too or have this improvement merged first?

ZippeyKeys12:
I can look at a more complete fix later, but merging this would unblock https://github.com/rems-project/cn/pull/111

kmemarian:
ok, merged.

PR #958, opened 2025-05-24
Less markers on ccall

kmemarian:


kmemarian:
@dimecon @cp526 if your ghost params branch was assuming `AilEcall` and `Eccall` always have markers you may need to tweak something.

PR #976, opened 2025-07-15
Add cerb byte attr

dc-mak:


dc-mak:
~~I'm using this to test various features etc. of the frontend (excl. execution, because currently that fails)~~

That's all working fine (tests in PR) but I'm struggling to figure out how to hit this specific line in a test:
https://github.com/rems-project/cerberus/blob/da57b9a1f15d5a0074acd6619ddaa59f2cdecfc2/frontend/model/translation.lem#L784

Here's a bunch of attempts that don't work/produce invalid Ail nodes

```c
[[cerb::byte]] typedef unsigned char byte;

// int global = 0;
// byte byte_arr1[] = { *((byte*)&global) };

struct s {
    int x;
    byte b;
};

int main()
{
    int x = 0;
    byte* p = (byte*)&x;
    byte p0 = p[0];
    byte arr[] = { p0 };
    // byte p[4] = { p0 };
    // struct s foo = { .x = x, .b = p0 };
    return arr[0];
}
```

PR #979, opened 2025-08-18
Fix web build for Byte

dc-mak:


dc-mak:
@kmemarian good to go?

PR #982, opened 2025-10-17
wrapI/catch-exceptional-condition for div/mod

cp526:
Apply same wrapI/catch-exceptional-condition handling to div/mod as to other arithmetic operations.


kmemarian (approved):
I appended some rework of the elaboration to your commit. I'll merge it once it passes the tests

Issue #443, opened 2024-07-27
[CN] Semantics of `unchanged` is confusing (eg. local vars, changed vars) 


septract:
(EDIT 2024-7-30: [see below](https://github.com/rems-project/cerberus/issues/443#issuecomment-2259312785) for some more commentary and thoughts about how we could make `unchanged` better) 

CN requires that loop invariants state which variables are unchanged, but this doesn't work for local variables. Eg it seems like this should work: 

```C
// Write to a local variable outside a loop, then use the value inside the loop. 
// Currently does not work. 

void loop_9 ()
{
  int test = 10; 
  for (int i = 0; i < 2; i++ ) 
  /*@ 
    inv 
      {test} unchanged; 
  @*/
  { 
    if (test != 10 ) {
      assert(0); 
    }
  }
}
```

... but it gives the following error message: 
```
$ cn verify loop_9.c
loop_9.c:10:8: error: Cannot resolve the value of &test in evaluation scope 'start'. Is the ownership for accessing &test missing?
      {test} unchanged; 
       ^
```

Going by the error message, perhaps `unchanged` is asserting that the value of the variable is the same as it was at function initialization? In that case, `test` has not been declared at this point.  

The following variant that constrains the value of `test` does pass CN:
```C
void loop_9_alt ()
{
  int test = 10; 
  for (int i = 0; i < 2; i++ ) 
  /*@ 
    inv 
      test == 10i32; 
  @*/
  { 
    if (test != 10 ) {
      assert(0); 
    }
  }
}
```

CN version: `git-b3d7e0463 [2024-07-26 17:41:55 -0700]`

septract:
This also does not verify: 

```C
void loop_10 (int test)
{
  test = 10; 

  for (int i = 0; i < 2; i++ ) 
  /*@ 
    inv 
      {test} unchanged; 
  @*/
  { 
    ; 
  }

  if (test != 10 ) {
    assert(0); 
  }
}
```

Error: 
```
$ cn verify inprogress/loop_10.c
[1/1]: loop_10
inprogress/loop_10.c:6:3: error: Unprovable constraint
  for (int i = 0; i < 2; i++ ) 
  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~ 
Constraint from inprogress/loop_10.c:9:7:
      {test} unchanged; 
      ^~~~~~~~~~~~~~~~~
```

peterohanley:
related, CN seems confused about the available resources:
```
[...]

PR #998, opened 2026-03-16
Mem2reg (Analysis + Transform plan)

dc-mak:
This PR adds support for the analysis part of a Core-to-Core mem2reg pass. This would be useful to speed up execution of Core programs and also CN proof (and Fulminate?) substantially.

It was developed with Claude and so I've included the plans that fed into the code changes for documentation in this PR - the final PR can have these deleted (including a fresh branch so they don't even exist in the history).

For example, Claude seemed to think that an Ail/Ail+Core analysis would be more complicated than a simple Core-to-Core, so the reasoning for this is included in doc/history/2026-03-13_mem2reg-plan.md

It's already got a few commits and a fair amount of code, but the PR can and should be read one commit at a time.

The analysis pass is quite naive - it does multiple traversals for one thing at a time (instead of one complex one to collect all the info), but it's easy to understand and I figured it can be improved later once we have a working implementation that passes tests (or not, if it turns out to be fast enough), e.g. I assume perhaps the Core rewriter library does some sort of pass-fusion?

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -5,9 +5,7 @@ Cerberus is a C semantics tool written in OCaml with a Dune build system.
 ## Build & run
 
 ```sh
-dune build                        # build everything
-dune build backend/driver/        # build the cerberus binary only
-./_build/default/backend/driver/cerberus --help
+make && make installl # buid everything and install it to the opam path
 
 # Common invocations
 cerberus file.c                   # parse + elaborate only
@@ -16,6 +14,8 @@ cerberus --pp core file.c         # pretty-print Core IR
 cerberus --typecheck-core file.c  # typecheck the Core IR
 cerberus --sw mem2reg file.c      # enable a switch
 ```
+
+Always use `make && make install` for building instead of dune commands.
 
 ## Architecture
 
@@ -61,10 +61,6 @@ automatically compiled into `cerb_frontend`.
        My_pass.transform_file core_file
      else core_file in
    ```
-
-## Buidling
-
-Always use `make && make install` for building instead of dune commands.
 
 ## Tests
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,161 @@
+# Cerberus
+
+Cerberus is a C semantics tool written in OCaml with a Dune build system.
+
+## Build & run
+
+```sh
+dune build                        # build everything
+dune build backend/driver/        # build the cerberus binary only
+./_build/default/backend/driver/cerberus --help
+
+# Common invocations
+cerberus file.c                   # parse + elaborate only
+cerberus --exec --batch file.c    # execute and print result
+cerberus --pp core file.c         # pretty-print Core IR
+cerberus --typecheck-core file.c  # typecheck the Core IR
+cerberus --sw mem2reg file.c      # enable a switch
+```
+
+## Architecture
+
+| Library / binary | Path | Role |
+|---|---|---|
+| `cerb_frontend` | `ocaml_frontend/` | C→AIL→Core elaboration, rewriters, switches |
+| `cerb_backend` | `backend/common/` | Pipeline orchestration (`pipeline.ml`) |
+| `cerberus` binary | `backend/driver/main.ml` | CLI entry point |
+
+The pipeline in `backend/common/pipeline.ml`:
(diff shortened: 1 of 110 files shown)


dc-mak:
Pasting this discussion here for visiblity.

**@cp526:** 
I suspect for `Esave` one wants to transform the label type. In the normal Core elaboration, for any in-scope C variable x, the `Esave` will take as an argument a pointer to the stack location of x. The mem2reg transformation should make it so the label instead takes the current value of x.

**@neel-krishnaswami:**
Operationally, `save` is basically just a let-binding, so it's easy. 

It's only when you hit a `run` that anything weird happens. Its operational rule uses the "labelmap" which (annoyingly) shows up in the run operational rule and nowhere else in Kayvan's thesis. Conceptually it's pretty straightforward but there may be details which live only in the Cerberus codebase and/or Kayvan's brain. 

**@dc-mak:** 
Ok so it seems like for the analysis, the default parameter of `save` counts as a use, but a slightly unusual one. As Neel pointed out, it is basically a let-binding, so we need to consider the binder as an alias _for analysing the body_. But for the transformation/executing the `save` as straightline code (not from a `run`), we would want it to count the default value as a `load`.  

AFAIR, elaboration ensures that `save`s are closed wrt C variables in scope (ie. only C local vars are bound in its binders). However, if this were not the case, any uses of a C local variable in the body (which were not bound by the `save` params), they would disqualify a local variable from promotion (since the `save` can be reached from straightline, forward run, and backward run, but any transformation would effectively lock-in the straightline case).

I think the fact that `save` parameters only bind the C local vars means that `run` should be easy. Unlike function calls, where passing the address of a local var would escape and disqualify it, passing the address of a local var to `run` would basically count as a `load`. I don't think labelmap will be much of a blocker, I remember understanding it once and I'll revisit it to be sure.

The transformation for `save` will be a little fiddly, like with `if`, since any mutated local vars will need to passed out the final expression, rebound with a `let strong`, and the environment updated.

**@cp526** For what it's worth, there's two alternatives to doing this as a core-to-core rewrite: option (1), which Kayvan was considering, is to adjust the elaboration around the variables mem2reg would optimise; option(2) would be to make this a mucore-to-mucore rewrite. We want to merge the core and mucore expression grammars, but will want to keep mucore's separate handling of labels (with labels as function-local procedures rather than Esave), so then you wouldn't have to think about how mem2reg should work for Esave

dc-mak:
**@cp526:**
For what it's worth, there's two alternatives to doing this as a core-to-core rewrite: option (1), which Kayvan was considering, is to adjust the elaboration around the variables mem2reg would optimise; option(2) would be to make this a mucore-to-mucore rewrite. We want to merge the core and mucore expression grammars, but will want to keep mucore's separate handling of labels (with labels as function-local procedures rather than Esave), so then you wouldn't have to think about how mem2reg should work for Esave

---

Other related points raised in discussion:
* Consider the impact on the code which supports reading local variables in at a program point within annotations (CN statements). We'd need to thread the environment of the transform through to that point.
* Consider that the promotable variables info could be useful to Fulminate; this may push towards an Ail-analysis at least, if not an Ail-transform too. IIRC, the symbols are the same in Ail and Core so this may not be a big concern.
* Another option is analysing at Ail/Core, but emulating the transform in the type system. This would make the type system more complicated.

dc-mak:
I've been coding and thinking some more and I'm now less sure about doing a mem2reg _transformation_ on Core. 
1. It can affect source locations.
2. The eliminated and new symbols will need to be threaded through to any backends like _anyways_ CN to support existing features to make sure that changes in mem2reg don't break specifications. The mem2reg pass would essentially need to preserve the environment used to support the ghost load in the middle of the function, or loop invariants.
    ```c
    int main() {
        int x = 10;
        while (x > 0)
        /*@ inv 0i32 < x; x <= 10i32 @*/
       { x--; }
       /*@ assert (x == 0i32); @*/
       return x;
     }
    ```

I always found the treatment of local vars in CN to be confusing (and I was not alone in this #443).

Could it be possible to end up in a situation where updating a mem2reg pass would cause existing verified code to break? That would be bonkers, but I am confused right now.

These considerations push me towards favouring an analysis and then passing those locations to CN, so they can be treated specially/on a fast path in the checker itself. But I am not decided/am still unsure.

PR #1001, opened 2026-03-30
Copy propagation

dc-mak:
These commits add a copy-propagation rewriter and switch to Core. It eliminates (replaces with wildcards and units) many redundant variable definitions and rebindings, which is useful for reading and debugging a Core program, and is necessary for other analyses like that for mem2reg.

The reason it does a replacement with wildcards/unit is (a) easier and (b) to keep any structure and annotations in place (except where necessary to to support CN or not lose location info precision).

Though each commit is self-contained, the code changed a lot during this process and so it's probably worth just reviewing it as a whole and squashing it rather than merging each commit (which can be kept for posterity).

I extracted these commits manually from the mem2reg branch/development, and tidied it up after to make it self-contained - I may have missed some stuff so do let me know and I can fix it.


ocaml_frontend/switches.ml
```diff
@@ -40,6 +40,9 @@ type cerb_switch =
   (* set magic comment syntax to "/*$ ... $*/" *)
   | SW_magic_comment_char_dollar
 
+  (* eliminate pure symbol rebindings: let alias = pure(sym) → substitute sym *)
```
dc-mak: Need to update this comment.

tests/run-copy-prop.sh
```diff
@@ -0,0 +1,106 @@
+#!/bin/bash
+# run-copy-prop-phase1.sh — regression check
+# Verifies that --sw copy_prop does not change output of existing CI tests (0001–0365).
```
dc-mak: Fix the numbering on this line.

tests/run-copy-prop.sh
```diff
@@ -0,0 +1,106 @@
...
+    echo -e "Test $file: \033[1m\033[33mMISSING .expected FILE\033[0m"
+    continue
+  fi
+
+  report "$file [+copy_prop]" "$file" "$ret"
+done
+
+echo ""
+echo "COPY_PROP PHASE 1 PASSED: $pass"
+echo "COPY_PROP PHASE 1 FAILED: $fail"
```
dc-mak: Remove "PHASE 1" from the printout.

dc-mak:
@kmemarian the pass is mostly independent from the rest of the code, but I did end up having to make some changes to `core_eval.lem` and `core_typing.lem` to so that regression testing with `run-copy-prop.sh` gave the same answer.

frontend/model/core_typing.lem
```diff
@@ -929,7 +929,16 @@ and typecheck_pexpr tagDefs (env: typing_env) (bTy: core_base_type) (Pexpr annot
...
+          | (Cspecified, BTy_storable, [pe]) ->
+              (* based on the infer_pexpr case for Cspecified, and the above *)
+              infer_pexpr tagDefs env pe >>= fun (Pexpr _ infer _ as pe') ->
+              match infer with
+                | InferredType_object oTy ->
+                    E.return (ret_ctor Cspecified pe')
+                | _ ->
+                    E.fail loc TooGeneral
+              end
+          | (Cspecified, _, pes) ->
```
dc-mak, line 943: More details on this: `tests/ci/0126-duff_device.c` that fails type checking without the proposed adjustment to the type checker.

Without copy prop, towards the end we have
```
      let strong (..., a_714: loaded pointer) =
        unseq(
          ...
        ,
          let weak a_715: pointer = pure(to{673}) in
          pure(Specified(array_shift(a_715, 'char', <@empty>:0)))
        ) in
      ...
          let strong a_708: pointer = pcall(create_and_store{27}, 'char*', a_714) in
```

With copy prop, we have
```
          let strong a_708: pointer =
            pcall(create_and_store{27}, 'char*', Specified(array_shift(to{673}, 'char', <@empty>:0))) in
```

Note that storable was created by Victor in commit a4ad3a93b2 and is only ever used once in std.core and std_inner_arg_temps.core

```
proc create_and_store (ty: ctype, x: storable) : eff pointer :=
  let weak p: pointer = create(Ivalignof(ty), ty) in
  let weak _: unit = store(ty, p, x) in
  pure(p)
```

And so without it we have the following error (modified for debugging)
```
unknown location  error: (Cspecified: Specified(array_shift(to{673}, 'char', <@empty>:0))):
this expression is of type 'loaded object' but an expression of type 'storable' was expected
```

PR #1004, opened 2026-05-24
(core) Add warning switch for wseq

dc-mak:


dc-mak:
Hmm, it seems a common way weak-sequencing may show up is with prefix incr/decr operators. It might be worth making the analysis more precise (e.g. taking into account promotable symbols can't alias and race), or it might just be that we need to bite the bullet and have it be supported by CN...

```c
#include <stdio.h>

int f() {
    int x = 0;
    int y = --x > x; // UB
    y = --x > y; // ok
    return 0;
}

int main() {
    f();
    int x[5] = {1,2,3,4,5};
    int *start = &x[0];
    int y = 5;
    do {
        printf("%d\n", x[--y]); // ok, rewritable
    } while (&x[y] > start);

    return x[y] - 5;
}
```

PR #1005, opened 2026-05-26
(core) Add jump/where expressions

dc-mak:
This series of commits adds two new constructors to Core, jump/where as a better-behaved, compositional alternative to run/save. jump is identical to run, but I've added it to keep the code simple.

A separate PR, to translate run/save Core programs to jump/where programs, will come later.

The semantics are described with just four rules, plus context reduction, ` C ::= .. | where(__, defs)`:

```
  [Where-Pure]
  ------------------------------
  pure(v) where defs --> pure(v)

  [Jump-Sub]
  ( defs(l) = x . E )
  ----------------------------------------------
  jump l(pe) where defs --> {val/x} E where defs

  [Jump-Where]
  (l not in defs)
  ------------------------------------
  jump l(pe) where defs --> jump l(pe)

  [Jump-Let]
  ---------------------------------------
  let strong pat = jump l(pe) in E --> jump l(pe)
```
Semantics are reminiscent of checked exceptions: Ejump
propagates out through Esseq continuations until caught
by an enclosing Ewhere.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -70,5 +70,6 @@ jobs:
         opam switch ${{ matrix.version }}
         eval $(opam env --switch=${{ matrix.version }})
         cd tests; USE_OPAM='' ./run-ci.sh
+        USE_OPAM='' ./run-save-to-where.sh
         ./diff-prog.py cerberus bytes/elab.json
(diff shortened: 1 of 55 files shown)


dc-mak:
FYI, this PR I made a special effort to have it be reviewable one commit at a time, so hope that helps.

PR #1006, opened 2026-05-29
(rewrite) const_prop: soundness & case support

dc-mak:
The previous analysis of which pexpr can be propagated was poorly explained and also unsound: values which could potentially not terminate, error, or cause UB, or depend on the memory model, in Core's strict setting should be left alone.

This shouldn't affect the upcoming mem2reg because that only relies on symbols being propagated correctly.

The PR also adds support for const-prop across case branches.


ocaml_frontend/rewriters/copy_propagation.ml
```diff
@@ -12,50 +12,58 @@ let extend_env_list env bindings =
...
+      | IntFromByte -> true
+    end && for_all Fun.id pes
+  | PEif _ | PEcase _ | PElet _ (* can be sound but inefficient *)
+  | PEbmc_assume _ (* for model checking only *)
+  | PEundef _ (* unsound to prop UB *)
+  | PEerror _ (* same for errors    *)
+  | PEcatch_exceptional_condition _ (* may raise UB *)
+  | PEconv_int _ (* may raise impl-defined error *)
+  | PEcfunction _ (* function pointer lookup may fail *)
+  | PEcall _ (* may not terminate *) ->
```
kmemarian: I dont see how `DeriveCap` can UB, it's not in the memory monad and the reduction in `Core_eval` simply returns the integer value.

Other than that, I double checked the rest and I think you have it all covered. Just some nitpick if we care about the comments being more precise:
*  `PEcall` may even also UB out.
*  for `PEcfunction`, looking at `Core_eval` I think all the `fail $ Illformed_program ...` should be UBs (there are also already a UB, and there will still remain an `error` for an illtyped expression, but that's irrelevant here). I'll make a PR to fix that in parallel.

kmemarian:
The CI failure is weird, I thought it was because the opam cache had an old cmdliner that might not do the argument shortening but it's using `2.1.0` and I can't replicate the issue with that version (or in fact older ones) on my machine.

ocaml_frontend/rewriters/copy_propagation.ml
dc-mak: So is it CapAssignValue that may UB here?
```ocaml
type pure_memop =
    (* CHERI *) (* (integer, integer) -> integer *)
  | DeriveCap of derivecap_op * bool (* is_signed *)
    (* CHERI*) (* (integer, integer) -> integer (may result in UB) *)
  | CapAssignValue
    (* CHERI *) (* integer -> integer *)
  | Ptr_tIntValue
    (* Bytes *)
  | ByteFromInt  (* integer -> byte *)
  | IntFromByte  (* byte -> integer *)
```

PR #1015, opened 2026-09-08
[DRAFT] Initialisers support

dc-mak:


frontend/model/cabs_to_ail.lem
```diff
@@ -2702,40 +2647,35 @@ and desugar_and_register_static_assert_declaration sa_decl =
...
 and desugar_declaration_base attrs specifs idecltors =
-  E.print_debugM 2 (fun () ->
-    "BROKEN: desugar_declaration_base"
-  ) >>= fun () ->
           desugar_type_specifiers specifs.type_specifiers >>= fun (specs_qs, base_ty) ->
+          let (isAtomic, tquals_qs) = translate_type_qualifiers specifs.type_qualifiers in
+          let base_qs = Ctype.combine_qualifiers specs_qs tquals_qs in
+          (* NOTE: `desugar_init_declarator' registers the identifiers it finds
+             (and takes care of isAtomic, of typedefs, and of the wf of the
+             storage classes). This is the same function used at file scope by
```
dc-mak, line 2657: [The comment to `desugar_init_declarator`](https://github.com/rems-project/cerberus/pull/1015/changes/e3f1bdaa8cf7b39f1358d9c10e530c7c98ad9790#diff-f446b8528b38196631037217233557454f51bc278a7eb73e235fc8213963f3a6L2155) says explicitly it doesn't handle wf of storage classes, so something is not consistent here.

frontend/model/cabs_to_ail.lem
```diff
@@ -2101,56 +2096,6 @@ STD_ "§6.7.1#8" $
...
-            (* NOTE: not using <$> because of a bug in Lem *)
-            wip_desugar_initializer_ loc (determinate_storage_duration link scs) ty init >>= fun z ->
-            E.return (Just z) in
-          E.return (ident, false, (qs, ty), mk_desug_init)
-  end
-
-
-
 and desugar_init_declarator attrs isAtomic has_alignment base_qs base_ty (scs: list storage_class_specifier) idecltor =
   (* NOTE: this function does NOT check the wf of the storage classes *)
```
dc-mak, line 2100: This comment.

frontend/model/cabs_to_ail.lem
dc-mak, line 1: @kmemarian This refactor needs your attention - beyond my familiarity.

dc-mak, line 1: Aside from that, the two claims in `e3f1bda` commit message are made without testing.
```
Two incidental consequences of sharing the file-scope function:
  - _Atomic block declarations now go through `wrap_atomic' rather than
    `Ctype.mk_ctype_atomic', so `_Atomic int *p;' inside a block behaves as
    it already did at file scope instead of hitting an internal error;
  - a typedef with an initialiser, or with an _Alignas, inside a block is
    now reported instead of silently ignoring the initialiser.
```

tests/ci/0343-init-designators.c
dc-mak, line 1:
```diff
diff --git a/tests/ci/0343-init-designators.c b/tests/ci/0343-init-designators.c
index 237210b36..82f4b524e 100644
--- a/tests/ci/0343-init-designators.c
+++ b/tests/ci/0343-init-designators.c
@@ -9,7 +9,7 @@ struct T { int a; int b; int c; };
 struct T t = {.b = 2, 3};

 // designators into an array, out of order, with a gap
-int a[5] = {[3] = 4, [1] = 2};
+int a[5] = {[3] = 4, 3, [1] = 2};

 // designators on an array of unknown size: the size comes from the largest
 // index reached, not from the number of initialisers
@@ -17,7 +17,7 @@ int unsized[] = {[4] = 5, 6};

 // nested designators
 struct U { struct T ts[2]; int n; };
-struct U u = {.ts[1].b = 7, .n = 9};
+struct U u = {.ts[1].b = 7, 8, .n = 9};
```

tests/ci/0344-init-anonymous-members.c
dc-mak, line 1: Is this affected by 6.7.9.9 - "Unnamed members of structure objects have indeterminate value even after initialization." or is that for something else?

tests/ci/0345-init-union-designators.c
```diff
@@ -0,0 +1,42 @@
...
+union U designated = {.b = 1.5};
+
+// re-designating: the last designator for the union wins
+union U redesignated = {.a = 1, .b = 2.5};
+
+// ... and the other way round
+union U redesignated2 = {.b = 2.5, .a = 3};
+
+// no designator: the first member
+union U implicit = {9};
```
dc-mak, line 21: Is assigning an a float to an int a constraint violation? What is/would be?

dc-mak:
Are `is_block` and `is_overflowing` independent? Presumable `<| is_block=false; is_overflowing=true; kind |>` is impossible/doesn't make sense?

```ocaml
type init_cursor_elem = <|
  is_block: bool; (* indicates whether an Init_list was entered *)
  is_overflowing: bool; (* indicates whether the cursor has overflowed (the current struct or array) *)
  kind: init_cursor_kind;
|>
```

dc-mak:
Desugaring_init.lem should really have a comment at the top of the file explaining the basic ideas for orientation, because it's elegant and simple but takes a while to grasp just by reading it.
1. Designators are interpreted/modelled using a zipper data structure into C types. They come with the usual operations for going up, down, next, but with some additions to handle brace elision and the rules of what's a valid designator.
2. First, the designator is iterated over (left to right), with the current position represented with a zipper, to construct a map. The keys of that maps are derived from the current position zipper. The values of that map are the expressions the user wrote.
3. Second, the value is constructed by traversing the C type recursively, keeping a track of the path to that point. That path is used to index the map from the previous step - if there's a value there, it's used, if not, the traversal recurses in the array/struct/union cases, bottoming out with just zeroes of the required type.

frontend/model/desugaring_init.lem
```diff
@@ -448,9 +448,9 @@ val     go_bottom_or_stop: map Symbol.sym tag_definition -> AilSyntax.expression
 let rec go_bottom_or_stop tagDefs d_e pos =
   let stops_here =
     match d_e with
-      | AilSyntax.AnnotatedExpression _ _ loc (AilSyntax.AilEstr (Nothing, strs)) ->
-          (* STD §6.7.9#14 *)
-          if AilTypesAux.is_array_of_character pos.ty && not (AilTypesAux.is_array_unknown_size pos.ty) then
+      | AilSyntax.AnnotatedExpression _ _ loc (AilSyntax.AilEstr ((_, strs) as lit)) ->
+          (* STD §6.7.9#14 and #15 *)
+          if array_is_initialisable_by_string_literal pos.ty lit && not (AilTypesAux.is_array_unknown_size pos.ty) then
```
dc-mak, line 453: Hmm, why does this do this if `cabs_to_ail.lem` explodes string?

tests/ci/0348-init-wide-string-literal.c
dc-mak, line 1: This should have some negative tests to get some constraint violations

frontend/model/ail/genTyping.lem
kmemarian, line 1: This is wrong. The point of `AilEinvalid` is to not get stuck on the frontend (this was added for soteria, but useful for others). The concern about the elaboration should be addressed by instead having a check for occurrence of this AST node before beginning the elab.

But this should not be part of this PR.

frontend/model/desugaring_init.lem
```diff
@@ -557,9 +561,15 @@ let rec interp_initializer_aux tagDefs funcs is_static_or_thread inits_acc is_in
...
         end
+      else if cursor_overflowing pos.cursor then
+        (* NOTE: the cursor has already run off the end of the object being
+           initialised, so this whole brace group is in excess: there is
+           nowhere to put it. We stay overflowing (STD §6.7.9#2). *)
+        E.return (max_index, inits_acc, true, pos)
       else
         (* NOTE: we go on step down (an mark the opening of a block { ... }) *)
         (* NOTE: this moves the cursor to the first member/index of the struct/array *)
+        let was_overflowing = cursor_overflowing pos.cursor in
```
dc-mak, line 572: Given this in the else branch we already know this is false?

frontend/model/desugaring_init.lem
```diff
@@ -591,9 +601,12 @@ let rec interp_initializer_aux tagDefs funcs is_static_or_thread inits_acc is_in
...
               (* NOTE: we move the cursor back to the level of the block we just opened *)
               go_top_of_block pos' >>= fun pos' ->
-              (* NOTE: we move up from this block *)
-              (* TODO: if we were overflowing before entering the block we need to STILL be overflowing *)
+              (* NOTE: we move up from this block. The overflow flag of the
+                 level we are leaving is about that block, not about the level
+                 we are returning to: if we were overflowing before entering
+                 the block, we must still be overflowing after leaving it. *)
               liftM (go_up (unmark_as_overflowing pos')) >>= fun pos' ->
+              let pos' = if was_overflowing then mark_as_overflowing pos' else pos' in
```
dc-mak, line 609: And so this is always pos' becuase `was_overflowing` is always false.

frontend/model/desugaring_init.lem
```diff
@@ -541,12 +541,16 @@ let rec interp_initializer_aux tagDefs funcs is_static_or_thread inits_acc is_in
...
              so [go_next] would descend into it. As with an Init_list, we move
              on ourselves and tell the caller not to advance again. *)
           liftM (do_next_in_same_level tagDefs pos) >>= fun pos ->
           E.return (max_index, inits_acc', true, pos)
+        else if whole_aggregate then
+          (* NOTE: an empty cursor means the whole object was initialised by
+             this one expression: there is no next position to move to. *)
+          E.return (max_index, inits_acc', true, pos)
         else
           E.return (max_index, inits_acc', false, pos)
```
dc-mak, line 555: [2a55eb2](https://github.com/rems-project/cerberus/commit/2a55eb2c0f7856c66826e21b11a707e23cbbbfe8) says "Finally, guard the \`do_next_in_same_level' call added in the previous commit against an empty cursor, which means the whole object was initialised by a single expression and there is no next position." but (a) it was added two commits prior and (b) the last two branches can be unified by saying `E.return (max_index, inits_acc', whole_aggregate, pos)`.

tests/ci/0350-init-excess-warning.c
```diff
@@ -0,0 +1,22 @@
+#include <stdio.h>
+
+// STD §6.7.9#2: "No initializer shall attempt to provide a value for an
+// object not contained within the entity being initialized." We accept these
+// (as GCC and Clang do) but warn; --switches=strict_initialisers turns the
+// warnings into constraint violations.
+//
+// The warnings go to stderr; this test checks that the values are dropped and
+// that compilation still succeeds.
```
dc-mak, line 9: The .expected file is about the stderr output of the C program itself, not the stderr warnings of Cerberus when run on those C programs, so this this test does not working as intended.

